PR #1, opened 2025-07-21
DSTEW-186: Chore/add db connectivity

vvartakMoJ:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/DSTEW-186)
Add ability to connect to the RDS (postgres) database

Unit tests, Integration tests and local deployments should continue using H2/test containers but main branch deployments (from UAT, staging or prod) should connect to the RDS database.

Secret from the K8s environment should be used to get the db connection details.

## Checklist

### DIFF
--- a/.github/workflows/build-test-deploy.yml
+++ b/.github/workflows/build-test-deploy.yml
@@ -6,7 +6,6 @@ on:
   push:
     branches:
       - main
-      - chore/add-db-connectivity
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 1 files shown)


.helm/data-claims-api/values/staging.yaml
```diff
@@ -12,21 +12,21 @@ image:
...
+  name: laa-data-claims-api-staging-service-account
 
 service:
   type: ClusterIP
   port: 8080
 
 ingress:
   annotations:
-    external-dns.alpha.kubernetes.io/set-identifier: "laa-data-stewardship-civil-claims-api-dsp-chart-laa-data-stewardship-civil-claims-api-staging-green"
+    external-dns.alpha.kubernetes.io/set-identifier: "laa-data-claims-api-dsp-chart-laa-data-claims-api-staging-green"
```
naseberry, line 23: Could we use the same chart name as the chart directory, i.e `data-claims-api` instead of `dsp-chart`

PR #3, opened 2025-07-22
DSTEW-186: add db connectivity

vvartakMoJ:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/DSTEW-186)

Add ability to connect to the RDS (postgres) database.

Unit tests, Integration tests and local deployments should continue using H2 in-memory database but main branch deployments (in UAT, staging or prod) should connect to the RDS database.

Secret from the K8s environment should be used to get the db connection details.

## Checklist


.github/actions/deploy_branch/action.yml
```diff
@@ -75,15 +75,25 @@ runs:
...
 
         echo "Deploying commit: $GIT_SHA under release name: $RELEASE_NAME to $release_host..."
+        
+        echo "branch name: $branch_name"
 
+        if [ "$branch_name" == "main" ]; then
+          export SPRING_PROFILE="main"
+        else
+          export SPRING_PROFILE="preview"
+        fi
```
naseberry, line 88: Remember, this file/action is only used when [deploying to uat](https://github.com/ministryofjustice/laa-data-claims-api/blob/main/.github/workflows/build-test-deploy.yml#L185), therefore would we need to replicate this on the `...deply/action.yml` as well?

PR #5, opened 2025-07-23
DSTEW-108: Add prod helm value and deploy step

naseberry:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/DSTEW-108)

1. Setup Production environment
2. Setup helm chart values fro Production.
3. Add Deployment steps for Production.

## Checklist


.github/workflows/build-test-deploy.yml
```diff
@@ -6,6 +6,7 @@ on:
   push:
     branches:
       - main
+      - dstew-108-production-deploy
```
vvartakMoJ, line 9: Assuming this was used during development/testing, can it be removed now before merge?

.helm/data-claims-api/values/production.yaml
```diff
@@ -0,0 +1,27 @@
...
+ingress:
+  annotations:
+    external-dns.alpha.kubernetes.io/set-identifier: "laa-data-claims-api-laa-data-claims-api-prod-green"
+    external-dns.alpha.kubernetes.io/aws-weight: "100"
+  hosts:
+    - laa-data-claims-api.cloud-platform.service.justice.gov.uk
+  tls:
+    - hosts:
+      - domain: laa-data-claims-api.cloud-platform.service.justice.gov.uk
+
```
vvartakMoJ, line 27: Due to changes for RDS connectivity (in my branch), you may have to add this setting here (unless we pass it in from the `helm upgrade` command in `deploy/action.yml`):

```
spring:
  profile: main
```

vvartakMoJ (dismissed):

.helm/data-claims-api/values/production.yaml
naseberry: Agreed to merge this PR first, and connectivity changes to be done in this [PR](https://github.com/ministryofjustice/laa-data-claims-api/pull/3)

vvartakMoJ (approved):

PR #15, opened 2025-07-31
Chore/dstew 228 local postgres

vvartakMoJ:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/DSTEW-228)
Replace the H2 in-memory database currently used in dev and for integration and unit testing with a postgres instance (running via docker compose for local dev and via testcontainers for integration and unit testing).
This resolves the problems caused by incompatibility between postgres-specific SQL scripts and the H2 database.

## Checklist


MapperAtMoJ (commented):
@vvartakMoJ just a few questions

PR #20, opened 2025-08-01
DSTEW-247: Implement javers

vvartakMoJ:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/DSTEW-247)

Implement javers to:
1. create the javers tables
2. store history of data changes in them.

## Checklist


.helm/data-claims-api/values/uat.yaml
```diff
@@ -22,6 +22,7 @@ ingress:
     nginx.ingress.kubernetes.io/enable-modsecurity: "true"
     nginx.ingress.kubernetes.io/modsecurity-snippet: |
       SecRuleEngine On
+      SecRuleRemoveById 911100
```
vvartakMoJ, line 25: Necessary to allow PUT and DELETE endpoints, not sure if this is the best way though.

.helm/data-claims-api/values/uat.yaml
```diff
@@ -22,6 +22,7 @@ ingress:
     nginx.ingress.kubernetes.io/enable-modsecurity: "true"
     nginx.ingress.kubernetes.io/modsecurity-snippet: |
       SecRuleEngine On
+      SecAction "id:900200,phase:1,nolog,pass,t:none,setvar:tx.allowed_methods=GET HEAD POST OPTIONS PUT PATCH DELETE"
```
vvartakMoJ, line 25: This is necessary to allow PUT, PATCH and DELETE methods which are disallowed by default.

PR #40, opened 2025-08-20
Added GET: /api/v0/submission/{id}/matter-starts/{matter-starts-id}

JamieBriggsDev:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/CCMSPUI-808)

GET: /api/v0/submission/{id}/matter-starts/{matter-starts-id}

## Checklist

### DIFF
--- a/claims-data/api/open-api-specification.yml
+++ b/claims-data/api/open-api-specification.yml
@@ -196,7 +196,7 @@ paths:
 
   /api/v0/submissions/{id}/matter-starts:
     post:
-      operationId: createMatterStart
+      operationId: getMatterStart
       tags:
         - Matter Starts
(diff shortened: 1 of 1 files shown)


claims-data/api/open-api-specification.yml
```diff
@@ -756,4 +794,21 @@ components:
...
+          type: string
+        procurement_area_code:
+          type: string
+        access_point_code:
+          type: string
+        delivery_location:
+          type: string
+        number_of_matter_starts:
+          type: integer
+          example: 3
```
MapperAtMoJ, line 814: Why are we re-introducing `number_of_matter_starts` here? I thought we said we were going to derive this and certainly not going to add to the `matter_start` record?

PR #46, opened 2025-08-26
Feature/dstew 339/implement open api spec for search endpoint

MapperAtMoJ:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/DSTEW-339)

Modified open-api spec to add get/submissions end point to filter claim submissions by office codes, submissionId, submittedDateFrom, submittedDateTo.

## Checklist

### DIFF
--- a/claims-data/api/open-api-specification.yml
+++ b/claims-data/api/open-api-specification.yml
@@ -235,7 +235,7 @@ paths:
 
       responses:
         '200':
-          description: list of paginated claim submissions (if any found)
+          description: returns a list of paginated claim submissions
           content:
             application/json:
               schema:


claims-data/api/open-api-specification.yml
```diff
@@ -364,8 +409,41 @@ paths:
                 type: object
                 description: submission details
                 properties:
-                  submission:
-                    $ref: '#/components/schemas/SubmissionFields'
+                  submission_id:
```
vvartakMoJ, line 412: What's the reason to not use SubmissionFields here? Assuming there is a good reason, is there any way to avoid duplicating these 30+ lines exactly here and in SubmissionFields later?

vvartakMoJ: Please try this:

allOf:
  - $ref: '#/components/schemas/SubmissionFields'
  
I hope it does the same thing (i.e. removes the extra wrapper "submission" : { ...} )

vvartakMoJ (dismissed):
LGTM!

claims-data/api/open-api-specification.yml
```diff
@@ -196,6 +196,51 @@ paths:
...
+          in: query
+          required: false
+          schema:
+            type: string
+            format: date
+          description: Filter submissions submitted up to this date (inclusive)
+
+      responses:
+        '200':
+          description: list of paginated claim submissions (if any found)
```
PhilDigitalJustice, line 238: Maybe just wording here: 

We always return an list, that list of content can be empty.
A search requests can be valid but no data is returned for that search.
Searches are always a bit more lenient than requests with path params.

PhilDigitalJustice (approved):
LGTM

PR #72, opened 2025-10-02
Fix: Add missing env variable

naseberry:
## What

Add `ECR_REGION` variable to deploy action as the the deploy action failed on staging and therefore production.

UAT used the deploy_branch action therefore this was already ok.

## Checklist


PhilDigitalJustice (approved):
LGTM

PR #73, opened 2025-10-03
Changes to allow authentication of office passed via request param

PhilDigitalJustice:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/DSTEW-560)

We need to implement office-level authorisation checks for bulk submission uploads to prevent users from submitting files for offices they are not authorised to access. This security control will validate that the office account number in the submitted file matches one of the offices the user is authorised to access.

Background

Currently, the bulk submission API (POST /api/v0/bulk-submissions) does not validate whether a user has the appropriate authorisation to submit claims for a specific office. This creates a security risk where users could potentially submit bulk submission files for offices they should not have access to.

Users will have a list of authorised offices associated with their account (as shown in the attached image with office codes like "0P322F", "2L849T", etc.). We need to validate submissions against this list before processing.

## Checklist


NagarjunaRachakonda (approved):
LGTM

PR #79, opened 2025-10-08
DSTEW-565: Fix v19 Flyway migration

naseberry:
## What

Add v20 migration to fix uat (main) CrashLoopBackOff status that seems to have been caused by V19:

- migration V19 tries to add a NOT NULL column provider_user_id to the submission table
- the column is added first with NOT NULL constraint
- then it tries to populate the column with data from bulk_submission table
- postgre fails because we cannot add a NOT NULL column to a table that already has data

## Checklist

### DIFF
--- a/claims-data/service/src/main/resources/db/migration/V20__fix_v19_provider_user_id_migration_error.sql
+++ b/claims-data/service/src/main/resources/db/migration/V20__fix_v19_provider_user_id_migration_error.sql
@@ -1,0 +1,42 @@
+-- V20: Fix the error from V19 migration
+-- V19 failed because it tried to add a NOT NULL column to a table with existing data
+-- This migration cleans up and properly adds the provider_user_id column
+
+-- Step 1: Remove the failed V19 migration from flyway history
+-- This allows migrations to continue past the failure
+DELETE FROM flyway_schema_history WHERE version = '19';
+
+-- Step 2: Check if column exists and handle appropriately
+-- Only add the column if it doesn't exist
+DO $$
+BEGIN
+    IF NOT EXISTS (
+        SELECT 1 FROM information_schema.columns 
+        WHERE table_schema = 'claims' 
+        AND table_name = 'submission' 
+        AND column_name = 'provider_user_id'
+    ) THEN
+        -- Column doesn't exist, safe to add it
+        ALTER TABLE submission ADD COLUMN provider_user_id TEXT;
+    ELSE
+        -- Column exists but might be NOT NULL (causing the original error)
+        -- Make it nullable first so we can populate it safely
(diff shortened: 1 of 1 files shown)


NagarjunaRachakonda (dismissed):

claims-data/service/src/main/resources/db/migration/V20__fix_v19_provider_user_id_migration_error.sql
```diff
@@ -0,0 +1,42 @@
+-- V20: Fix the error from V19 migration
+-- V19 failed because it tried to add a NOT NULL column to a table with existing data
```
MapperAtMoJ, line 2: I think it's more like "it tried to add a NOT NULL constraint to a column before this was pre-populated" (Step 4 and 5 below)

MapperAtMoJ (commented):
LGTM. Could you please update this [table](https://dsdmoj.atlassian.net/wiki/spaces/WHCZA/pages/5866160240/Flyway+migration+history) as well @naseberry? Thank you!

naseberry:
> If V19 didn't run we should amend this and re-deploy instead of creating V20.

@MapperAtMoJ  consider v19 worked on stg and prod, that has no data, I thought it logical to create a v20 to resolve all environments.

claims-data/service/src/main/resources/db/migration/V20__fix_v19_provider_user_id_migration_error.sql
```diff
@@ -0,0 +1,42 @@
...
+-- This migration cleans up and properly adds the provider_user_id column
+
+-- Step 1: Remove the failed V19 migration from flyway history
+-- This allows migrations to continue past the failure
+DELETE FROM flyway_schema_history WHERE version = '19';
```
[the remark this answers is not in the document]
naseberry: See second commit

NagarjunaRachakonda (approved):

NagarjunaRachakonda (approved):

claims-data/service/src/main/resources/db/migration/V20__fix_v19_provider_user_id_migration_error.sql
```diff
@@ -1,42 +1,67 @@
...
+        SET provider_user_id = (
+            SELECT bs.created_by_user_id
+            FROM bulk_submission bs
+            WHERE bs.id = s.bulk_submission_id
+        );
+        
+        -- Handle nulls
+        UPDATE submission 
+        SET provider_user_id = 'unknown_user' 
+        WHERE provider_user_id IS NULL;
```
MapperAtMoJ, line 60: same as above.

MapperAtMoJ (commented):
@naseberry just added a couple of comments.

PR #300, opened 2026-04-01
[DSTEW-1414] - Support partial and case-insensitive CRN search

BenJusticeDigital:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/DSTEW-1414)

This PR implements the database and code changes required to support faster, case-insensitive substring searches of case reference numbers, plus a few housekeeping/quality fixes and tests. 

###  SQL (performance / migration)
- Enable the pg_trgm extension and add a GIN trigram index on lower(case_reference_number) to accelerate case-insensitive contains queries (LOWER(case_reference_number) LIKE '%term%').

###  RequestValidator
- Created a validator to centralise the logic when receiving ClaimSearchRequest
- Added 3 char check to claim reference number

###  DataNormaliser
- Made the class a proper utility class by adding a private constructor (prevents instantiation).
- Kept normaliseClaimSearchRequest(...) semantics: trims and converts whitespace-only values to null.

###  Tests
- Added parameterized tests covering /api/v2/claims case_reference behaviour:
- Partial match (e.g., CRN ABC-1234 searched by ABC)
- Contains match (e.g., RAC ATE2/1 searched by ATE2/1)
- Case-insensitive match (ATE2/1 vs ate2/1)
- Short CRN rejection (enforces the min-3 char rule)
- No-match, v1 behaviour unchanged, and pagination/filters unaffected
- These tests create a submission + claim, then exercise the v2 endpoint and assert expected results.
- Fixed a Mockito stubbing issue by returning a mocked Path<String> from root.get(...) (instead of an Expression) to avoid 

## Checklist


claims-data/service/src/test/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/ClaimServiceTest.java
```diff
@@ -672,6 +697,77 @@ void getClaimResultSet_v2_whenFiltersMatchData_shouldReturnNonEmptyResultSet() {
...
+            .claimStatuses(List.of(ClaimStatus.READY_TO_PROCESS))
+            .submissionPeriod(SUBMISSION_PERIOD)
+            .caseReferenceNumber(caseRef)
+            .build();
+
+    if (shouldThrow) {
+      ClaimBadRequestException ex =
+          assertThrows(
+              ClaimBadRequestException.class,
+              () -> claimService.getClaimResultSetV2(request, Pageable.unpaged()));
```
andrewJohnysJustice: There's two invocations inside the assertThrows

andrewJohnysJustice (commented):
Just the Deprecated Javadoc and the assertThrows - looks great

andrewJohnysJustice (approved):

vvartakMoJ (approved):
LGTM!

bruno/collection.bru
```diff
@@ -0,0 +1,3 @@
+headers {
+  Authorization: {{AUTH_TOKEN}}
```
vvartakMoJ, line 2: With this in place now (and inherited from here by other files) is it worth removing the `Authorization: {{AUTH_TOKEN}}` sections from the V1 files too?

BenJusticeDigital: Yes.. but i think out of scope for this pull request.

PR #316, opened 2026-04-14
[BC-90] - Return empty assessment set

mboyd-moj:
## What

If a claim exists but has no assessments then we should return a 200 and an empty result set instead of a 404

## Checklist

### DIFF
--- a/claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/AssessmentService.java
+++ b/claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/AssessmentService.java
@@ -1,5 +1,7 @@
 package uk.gov.justice.laa.dstew.payments.claimsdata.service;
 
+import static uk.gov.justice.laa.dstew.payments.claimsdata.service.ClaimValidationService.NO_CLAIM_FOUND_WITH_ID_ERROR;
+
 import jakarta.validation.constraints.NotNull;
 import java.math.BigDecimal;
 import java.util.List;
@@ -16,6 +18,7 @@ import uk.gov.justice.laa.dstew.payments.claimsdata.entity.Assessment;
 import uk.gov.justice.laa.dstew.payments.claimsdata.entity.Claim;
 import uk.gov.justice.laa.dstew.payments.claimsdata.entity.ClaimSummaryFee;
 import uk.gov.justice.laa.dstew.payments.claimsdata.exception.AssessmentNotFoundException;
+import uk.gov.justice.laa.dstew.payments.claimsdata.exception.ClaimNotFoundException;
 import uk.gov.justice.laa.dstew.payments.claimsdata.mapper.AssessmentMapper;
 import uk.gov.justice.laa.dstew.payments.claimsdata.model.AssessmentGet;
 import uk.gov.justice.laa.dstew.payments.claimsdata.model.AssessmentPost;
@@ -123,22 +126,21 @@ public class AssessmentService {
    * <ul>
    *   <li>Validates that the provided {@code claimId} is not null.
    *   <li>Fetches assessments from the repository ordered by creation date (descending).
-   *   <li>If no assessments are found, throws {@link AssessmentNotFoundException}.
    *   <li>Maps the list of assessments to an {@link AssessmentResultSet} using the mapper.
    * </ul>
    *
    * @param claimId the unique identifier of the claim; must not be {@code null}.
    * @return an {@link AssessmentResultSet} containing all assessments for the claim.
    * @throws IllegalArgumentException if {@code claimId} is {@code null}.
-   * @throws AssessmentNotFoundException if no assessments exist for the given claim ID.
+   * @throws ClaimNotFoundException if no claim exists for the given claim ID.
    */
   @Transactional(readOnly = true)
   public AssessmentResultSet getAssessmentsByClaimId(@NotNull UUID claimId, Pageable pageable) {
-    var assessments = assessmentRepository.findByClaimId(claimId, pageable);
-    if (assessments.isEmpty()) {
-      throw new AssessmentNotFoundException(
-          String.format("No assessments found for claimId: %s", claimId));
+    if (!claimRepository.existsById(claimId)) {
+      throw new ClaimNotFoundException(String.format(NO_CLAIM_FOUND_WITH_ID_ERROR, claimId));
     }
+
+    var assessments = assessmentRepository.findByClaimId(claimId, pageable);
     return assessmentMapper.toAssessmentResultSet(assessments);
   }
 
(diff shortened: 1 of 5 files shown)


claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/AssessmentService.java
```diff
@@ -123,22 +123,16 @@ public AssessmentGet getAssessment(UUID claimId, UUID assessmentId) {
...
    * @throws IllegalArgumentException if {@code claimId} is {@code null}.
-   * @throws AssessmentNotFoundException if no assessments exist for the given claim ID.
    */
   @Transactional(readOnly = true)
   public AssessmentResultSet getAssessmentsByClaimId(@NotNull UUID claimId, Pageable pageable) {
     var assessments = assessmentRepository.findByClaimId(claimId, pageable);
-    if (assessments.isEmpty()) {
-      throw new AssessmentNotFoundException(
-          String.format("No assessments found for claimId: %s", claimId));
-    }
```
naseberry, line 141: Now, should we update the OpenAPI spec to clarify that 404 is only returned if the endpoint itself is invalid, not for empty results? i.e. from `Claim not found` to `Invalid claim ID format or endpoint not found`

claims-data/service/src/test/java/uk/gov/justice/laa/dstew/payments/claimsdata/controller/AssessmentControllerTest.java
```diff
@@ -277,19 +276,6 @@ void shouldReturnAssessmentsForValidClaimId() throws Exception {
...
-
-      when(assessmentService.getAssessmentsByClaimId(eq(claimId), any(Pageable.class)))
-          .thenThrow(new AssessmentNotFoundException("No assessments found"));
-
-      mockMvc
-          .perform(get("/api/v1/claims/{claimId}/assessments", claimId))
-          .andExpect(status().isNotFound())
-          .andExpect(jsonPath("$.message").value("No assessments found"));
-    }
-
```
naseberry, line 292: Do we have/need a test to cover a `no assessment` scenario?

claims-data/service/src/test/java/uk/gov/justice/laa/dstew/payments/claimsdata/mapper/AssessmentMapperTest.java
```diff
@@ -151,6 +151,19 @@ void toAssessmentResultSet_mapsContentAndPagination() {
...
+  void toAssessmentResultSet_mapsEmptyPage() {
+    var page = new PageImpl<Assessment>(List.of());
+
+    var result = mapper.toAssessmentResultSet(page);
+
+    assertNotNull(result);
+    assertEquals(0, result.getNumber());
+    assertEquals(0, result.getSize());
+    assertEquals(0, result.getTotalElements());
+    assertEquals(1, page.getTotalPages());
```
naseberry, line 164: Is this correct? If no assessment, shouldn't  getTotalPages be `0` instead of `1`?

mboyd-moj: Looks like this comes from the Spring Page implementation:
```
	@Override
	public int getTotalPages() {
		return getSize() == 0 ? 1 : (int) Math.ceil((double) total / (double) getSize());
	}
```
I guess it makes sense to define like this as one page of zero size, rather than zero pages of arbitrary size

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/AssessmentService.java
mboyd-moj: I don't believe there is any usecase now where we can get 404 here anymore other than mistyped URL which is true of any endpoint. An invalid claim ID gets raised as a 400. Should I just remove 404 from the spec?

claims-data/service/src/test/java/uk/gov/justice/laa/dstew/payments/claimsdata/controller/AssessmentControllerTest.java
mboyd-moj: I removed it here as there is now no functional difference in the Controller between a response with/without assessments, we just pass through whatever the service returned

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/AssessmentService.java
BenJusticeDigital: Yes if it is not a possible return value on this endpoint and is specified as one.

claims-data/service/src/test/java/uk/gov/justice/laa/dstew/payments/claimsdata/controller/AssessmentControllerTest.java
BenJusticeDigital: Although there’s no longer a functional difference in the controller logic, the test still serves to document and verify expected behaviour for an empty service response, rather than branching logic. So i believe we should still have a test for it.

BenJusticeDigital (changes requested):

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/AssessmentService.java
mboyd-moj: Removed

claims-data/service/src/test/java/uk/gov/justice/laa/dstew/payments/claimsdata/controller/AssessmentControllerTest.java
mboyd-moj: Sure, added it back in

claims-data/service/src/test/java/uk/gov/justice/laa/dstew/payments/claimsdata/mapper/AssessmentMapperTest.java
PhilDigitalJustice: Yeah spring always has a single page size as you have returned a payload

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/AssessmentService.java
mboyd-moj: Popped this back in as we now have 404 for missing claim

BenJusticeDigital (approved):
LGTM!

PR #319, opened 2026-04-15
[DSTEW-1538] - Fix bug in calculated total amount and ensure two decimal places even when ZERO 

andrewJohnysJustice:
## What

https://dsdmoj.atlassian.net/browse/DSTEW-XXX)](https://dsdmoj.atlassian.net/browse/DSTEW-1538

Fixed bug with calculated_total_amount and all values that are rendered by calling the BigDecimalUtil - to display 2 decimal places in all cases (except null)

## Checklist

### DIFF
--- a/claims-data/api/open-api-specification.yml
+++ b/claims-data/api/open-api-specification.yml
@@ -1304,10 +1304,15 @@ components:
           description: Number of claims in this submission.
         calculated_total_amount:
           type: number
+          format: decimal
           description: Total amount calculated from the fee details of each claim.
+          example: "0.00"
+          default: "0.00"
         assessed_total_amount:
           type: number
+          format: decimal
           description: Total amount calculated from the latest assessment for each claim.
+          example: "0.00"
         submitted:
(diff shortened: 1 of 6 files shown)


claims-data/api/open-api-specification.yml
```diff
@@ -1304,10 +1304,15 @@ components:
           description: Number of claims in this submission.
         calculated_total_amount:
           type: number
+          format: decimal
           description: Total amount calculated from the fee details of each claim.
+          example: "0.00"
+          default: "0.00"
```
naseberry: Is this a response from FSP, should we be setting a default?

andrewJohnysJustice: I've removed the default as discussed - we don't really need it

BenJusticeDigital (approved):
LGTM!

PhilDigitalJustice (approved):
LGTM

vvartakMoJ (approved):
LGTM

PR #321, opened 2026-04-21
DSTEW-1568: Disable Cronjobs on Ephemeral UAT

naseberry:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/DSTEW-1568)

Implement configuration to disable replication summary cron jobs on ephemeral UAT preview environments while maintaining functionality for persistent environments (main UAT, staging, and production), it was redundant (and noisy) to have these jobs running in this environment.

This change wraps the entire CronJob resource with a conditional check to enable/disable creation.

## Checklist

### DIFF
--- a/.helm/data-claims-api/values/production.yaml
+++ b/.helm/data-claims-api/values/production.yaml
@@ -61,3 +61,7 @@ sentry:
 
 portForward:
   enabled: false
+
+replicationSummary:
+  enabled: true
+  schedule: "0 1 * * *"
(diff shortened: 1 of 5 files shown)


.helm/data-claims-api/values/production.yaml
```diff
@@ -61,3 +61,7 @@ sentry:
 
 portForward:
   enabled: false
+
+replicationSummary:
+  enabled: true
+  schedule: "0 1 * * *"
```
naseberry, line 67: not strictly necessary since the template has a default, but makes the config more explicit

BenJusticeDigital (approved):

BenJusticeDigital (approved):

PR #328, opened 2026-04-28
DSTEW-472 reversion - Added V2 Endpoints

andrewJohnysJustice:
## What

https://dsdmoj.atlassian.net/issues?filter=-1&selectedIssue=DSTEW-1472

Append Latest Assessment Fields to Provider CSV Exports - when there are no assessments use defaults values from CFD to populate Final Claim Value and Final Claim Value VAT. Legal Help has 2 extra fields.

Add new features as Version 2 endpoints

## Checklist


PhilDigitalJustice:
Is there a reason for having as v2 it would be great not to have to update the endpoint for the UI. SaBC is currently the only consumers of these export endpoints. 

Changing the endpoints also means updating the authentications secrets?

PhilDigitalJustice:
Pact tests will also break due to the unversioned changes as you are performing a break in change for the export endpoints.

andrewJohnysJustice:
I'll talk to Peter about this, this morning - I hear what you are saying

PR #330, opened 2026-04-28
[DSTEW-1538] - fixed previous bug in Calc Total and added filter to the other endpoint as well.

andrewJohnysJustice:
## What

https://dsdmoj.atlassian.net/browse/DSTEW-1538

The calculated_total_amount was set to be ZERO or a Decimal back in October. Never null. This is incorrect. I have reversed this fix and added the new "Scale or Nullable" filter on Both endpoints as /api/v1/getSubmissions was ignoring the field completely. 

Assessment_Total_Value remains unchanged as this is correct for both endpoints.

## Checklist


BenJusticeDigital (approved):

vvartakMoJ (approved):
LGTM!

PR #337, opened 2026-05-15
Feature/[DSTEW-1527] - emit submission validation sucess event

g-georgieva:
## What

[Link to story](https://dsdmoj.atlassian.net/jira/software/c/projects/BC/boards/15952?selectedIssue=DSTEW-1527)

Amended claims-api to emit validation success event

## Checklist


claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/exception/SubmissionValidationSucceededQueuePublishException.java
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Exception for issues when attempting to publish a submission validation succeeded event.
+ *
+ * <p>This exception indicates an internal failure when publishing the submission validation
+ * succeeded event to the SNS topic. It extends {@link ClaimsDataException} and will automatically
+ * result in a {@link org.springframework.http.HttpStatus#INTERNAL_SERVER_ERROR 500} response being
+ * returned to the client.
+ */
+public class SubmissionValidationSucceededQueuePublishException extends ClaimsDataException {
```
BenMillar-MOJ: Would it be better to have a generic `QueuePublishException`, rather than an exception for this specific event failing?

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/SubmissionEventPublisherService.java
```diff
@@ -72,6 +73,28 @@ public void publishSubmissionValidationEvent(UUID submissionId) {
...
+   * @param submissionId the unique identifier for the submission
+   */
+  public void publishSubmissionValidationSucceededEvent(UUID submissionId) {
+    SubmissionValidationMessage submissionValidationSucceededMessage =
+        new SubmissionValidationMessage(submissionId);
+    try {
+      publishEvent(
+          submissionValidationSucceededMessage,
+          SubmissionEventType.SUBMISSION_VALIDATION_SUCCEEDED);
+    } catch (JsonProcessingException e) {
```
BenJusticeDigital: Is this the only exception this code can throw? Would it be preferable to take a more defensive approach and catch additional exceptions to prevent them from propagating further up the call stack?

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/SubmissionEventPublisherService.java
```diff
@@ -72,6 +73,28 @@ public void publishSubmissionValidationEvent(UUID submissionId) {
...
+      throw new SubmissionValidationSucceededQueuePublishException(
+          "Error when creating validation succeeded message for submission id ["
```
BenMillar-MOJ: You are throwing a `SubmissionValidationSucceededQueuePublishException`, however, the only exceptions which are raised by `publishEvent` are `JsonProcessingException`.

The exception you are raising suggests an issue with SQS, however, it seems like it would be more truthful to allow the `JsonProcessingException` to be thrown rather than catching it.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/SubmissionService.java
```diff
@@ -135,6 +135,11 @@ public void updateSubmission(UUID id, SubmissionPatch submissionPatch) {
       TransactionalPublisher.runAfterCommit(
           () ->
               submissionEventPublisherService.publishSubmissionValidationEvent(submission.getId()));
+    } else if (submissionPatch.getStatus() == SubmissionStatus.VALIDATION_SUCCEEDED) {
+      TransactionalPublisher.runAfterCommit(
+          () ->
+              submissionEventPublisherService.publishSubmissionValidationSucceededEvent(
```
BenMillar-MOJ, line 141: What value do you get from calling `publishSubmissionValidationSucceededEvent ` here rather than calling `SubmissionEventPublisherService.publishEvent` with the type of your event?

BenMillar-MOJ (changes requested):

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/SubmissionEventPublisherService.java
```diff
@@ -72,6 +73,28 @@ public void publishSubmissionValidationEvent(UUID submissionId) {
...
+   */
+  public void publishSubmissionValidationSucceededEvent(UUID submissionId) {
+    SubmissionValidationMessage submissionValidationSucceededMessage =
+        new SubmissionValidationMessage(submissionId);
+    try {
+      publishEvent(
+          submissionValidationSucceededMessage,
+          SubmissionEventType.SUBMISSION_VALIDATION_SUCCEEDED);
+    } catch (JsonProcessingException e) {
+      throw new SubmissionValidationSucceededQueuePublishException(
```
BenJusticeDigital: Do we want this scenario to result in a 500 error being returned to the calling service?
For example, if the Events service receives a 500 response, it would assume the update has failed and propagate that failure back to SQS, marking the entire validation process as unsuccessful, even though the validation itself may have passed.
Should we allow this behaviour, or would it be more appropriate to log the error and monitor it (e.g. via Sentry) without surfacing it as a failure to downstream systems?

claims-data/service/src/test/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/SubmissionEventPublisherServiceTest.java
```diff
@@ -107,4 +107,35 @@ void publish_ValidateSubmissionEvent_sendsMessageWithCorrectPayload() {
 
     verifyNoMoreInteractions(snsClient);
   }
+
+  @Test
+  void publish_ValidationSucceededEvent_sendsMessageWithCorrectPayload() {
```
BenJusticeDigital, line 112: This tests the happy path, lets make sure we cover what happens when it fails.

BenJusticeDigital (changes requested):

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/SubmissionEventPublisherService.java
g-georgieva: Good point, I have updated the method to log failures rather than propagating them and not just JsonProcessingException

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/SubmissionEventPublisherService.java
g-georgieva: Yes, that is a very good point, we shouldn't fail the overall validation flow, it does seem like a better approach to log the error instead and monitor the failure. The change I made for the previous comment should address that one I believe

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/exception/SubmissionValidationSucceededQueuePublishException.java
g-georgieva: The exception is no longer needed in general now because I changed it to log failures rather than propagate them as exceptions

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/SubmissionEventPublisherService.java
g-georgieva: Agreed, the exception naming.handling was  a bit misleading so I have simplified the flow now to log and not propagate exceptions

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/SubmissionService.java
g-georgieva: Originally this method was indeed very similar to the existing publish methods in the SubmissionEventPublisherService and I created a new method to follow the existing pattern of the repo as well as the description on the ticket. I agree back then I could have reused an existing one or just call publishEvent everywhere with a generic exception but now the publishSubmissionValidationSucceededEvent method logs and does not throw exceptions as opposed to the existing methods/pattern so it makes more sense to keep it as is but let me know what you think

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/exception/SubmissionValidationSucceededQueuePublishException.java
BenMillar-MOJ: Can this class be removed if it's no longer used

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/SubmissionService.java
BenMillar-MOJ: Yeah, this is much better. I think this area could do with a refactor, however, this PR looks good!

BenMillar-MOJ (changes requested):
Looks good, if we can remove the unused exception class it's good to go!

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/exception/SubmissionValidationSucceededQueuePublishException.java
g-georgieva: Yes,sorry I thought I had removed it but it wasn't pushed up

BenMillar-MOJ (approved):
Thank you!

BenJusticeDigital (approved):

PR #339, opened 2026-05-20
[DSTEW-1659] - Database changes for Claims Amendments

andrewJohnysJustice:
## What

https://dsdmoj.atlassian.net/browse/DSTEW-1659

**Production Code**

Database Migration & Schema Evolution (Flyway V39 & V40)
1. Introduction of the Claim Amendment Storage Model (V39)
What: Created the new claims.claim_amendment table and established its structural foreign key relationships with the existing claims.claim and claims.amendment_reason_reference tables.

Why: This isolates the amendment audit trail from the core transaction engine. Rather than altering a claim in place and losing its historical state, every modification creates an immutable snapshot anchor. This fulfills strict data auditing invariants by tracking exactly when an amendment happened, why it happened, and who authorized it.

2. Mapping Ledger History via Foreign Key Realignment (V39)
What: Refactored the claims.calculated_fee_detail table by adding a nullable claim_amendment_id foreign key column pointing back to the new amendment table.

Why: This enables the "Latest Wins" calculation ledger strategy. Initial submissions store calculation rows where claim_amendment_id is null. When an amendment occurs, the calculation engine runs again and flags the newly calculated rows with the corresponding claim_amendment_id. This preserves a perfect lineage of how fees changed over time while keeping queries fast.

3. Native JSONB Transport Storage Matrix (V39)
What: Defined the audit logging snapshot fields (before_state, request_payload, and diff) as native PostgreSQL jsonb data types.

Why: Traditional text or varchar columns risk truncation errors or indexing bottlenecks when dealing with massive, deeply nested data payloads from bulk submissions. Using jsonb ensures that Postgres natively decompresses, validates, and stores these string graphs with complete structural integrity while allowing for optimized JSON path queries if we need them down the line.

4. Standardized Global Reference Seeding (V40)
What: Implemented a native Postgres reference data seed script using Flyway to establish system constants like metadata reference tracking codes (PROVIDER, AUDITOR) and reason rules (TYPING_ERROR, COMPLIANCE_CORRECTION).

Why: Hardcoding lookup parameters in application code creates data drift across environments. By managing the seed state natively via Flyway using PostgreSQL's standard uuidv7() generation, local Testcontainers, staging clusters, and production environments are guaranteed to evaluate foreign key metadata lookups identically.

5. Version-Based Concurrency Controls
What: Enforced an integer-based locking tracking strategy (JPA @Version) mapped directly to the version column on the parent claims.claim table.

Why: In high-concurrency environments like legal aid processing, multiple background threads or administrative users might attempt to review or amend the same claim concurrently. Bumping this database version counter automatically whenever a direct property changes—or when a child relationship is appended—effectively isolates concurrent edits. Stale modifications are instantly blocked at the database boundary, completely eliminating data overwrite risks.

**For Testing...**
We have established a robust, production-grade integration test suite for the database persistence layer (ClaimAmendmentStorageIntegrationTest.java) utilizing Spring Boot integration tests running against PostgreSQL 18.3 via Testcontainers.

The suite thoroughly validates four critical database behavior domains:

Metadata Referential Integrity: Validates that ClaimAmendment entries strictly enforce foreign key constraints, instantly trapping phantom or un-seeded reference codes.

Ledger Tracking Invariants ("Latest Wins"): Confirms that our custom database collection fetching logic correctly processes multiple historical fee calculations, applying temporal ordering rules via Hibernate to surface the latest valid state.

Concurrency Control (Optimistic Locking): Proves that concurrent modifications to identical Claim snapshots trigger standard JPA optimistic locking mechanisms to eliminate data overwrite risks.

Heavy Payload Transport Channel Stability: Stress-tests the PostgreSQL native jsonb data transport layer to ensure heavily nested data strings are successfully streamed, written, and read back without silent truncation.

Why Have We Done It? (Technical Rationales)
Synchronizing Local Containers with Higher Environments
Instead of relying on programmatic database mocking inside our test setup blocks—which was creating duplicate key conflicts (uq_requested_by_reference_code) with baseline migration records—we refactored the suite to directly reference baseline reference data (PROVIDER, AUDITOR) seeded dynamically by Flyway migrations (V40). This ensures local test runs match the behavior of upstream staging and production database containers.

Navigating Hibernate First-Level Cache Boundaries
During full transaction updates, Hibernate reuses existing entity structures from memory. We introduced precise bi-directional cache mapping synchronization steps within active @Transactional boundaries. This forces the entity graph to accurately see child fee calculations and accurately verify ledger states without cache leakage.

Preventing State Overwrite via Version Counters
We validated that mutating properties across multiple repositories (e.g., child log insertions vs. direct parent field updates) correctly steps the @Version block incrementally. This ensures our concurrency defenses work seamlessly as a cohesive unit across multi-table operations.

Adhering to Standardized Test Assertions
We cleaned up legacy test constructs by swapping out standard JUnit assertion blocks for a fully fluent AssertJ assertion model. Additionally, temporal checking parameters were modernized from deprecated truncation operations to use explicit time-window testing bounds (isCloseTo(...) with a 1-second delta), rendering our specs entirely safe against microsecond/nanosecond rounding behaviors inherent to databases.

## Checklist


claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/repository/AmendmentReasonReferenceRepository.java
```diff
@@ -0,0 +1,33 @@
...
+
+  /**
+   * Fetch all valid, active amendment reasons scoped to a specific requested_by_reference row ID.
+   * Useful for building the dependent dropdown lists in the UI.
+   */
+  List<AmendmentReasonReference> findByRequestedByIdAndIsActiveTrueOrderByDisplayOrderAsc(
+      UUID requestedById);
+
+  /** Look up a reason by its unique composite constraint pair. */
+  Optional<AmendmentReasonReference> findByRequestedByCodeAndCode(
```
vvartakMoJ: Should "AndIsActiveTrue" be added to this method also, to stop non-active/retired reasons being returned?

claims-data/service/src/main/resources/db/migration/V39__claim_amendment_storage_model.sql
```diff
@@ -0,0 +1,105 @@
+-- -------------------------------------------------------------------
+-- 1. Metadata: Requested By Reference
+-- -------------------------------------------------------------------
+CREATE TABLE claims.requested_by_reference (
```
vvartakMoJ: We need to know/decide if any of these new tables are needed for reporting. If so then we will have to coordinate with the GLAD team and add it to the publication so that they get replicated.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/entity/ClaimAmendment.java
```diff
@@ -0,0 +1,60 @@
...
+/** Entity representing a claim amendment. */
+@Getter
+@Setter
+@Entity
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+@Table(name = "claim_amendment", schema = "claims")
+public class ClaimAmendment {
+  @Id private UUID id; // UUIDv7 [cite: 519]
```
vvartakMoJ: What do all these numbers in the comments (cite: 519) refer to? Can we add a note saying how to use them, or remove them if not useful, to avoid noise?

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/entity/AmendmentReasonReference.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.UUID;
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+
+/** Entity representing an amendment reason reference. */
+@Getter
+@Setter
```
vvartakMoJ: Can we safely remove @ Setter annotation here as it seems to be an immutable entity? Same for the other reference entity.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/entity/CalculatedFeeDetail.java
```diff
@@ -119,4 +120,10 @@ public class CalculatedFeeDetail {
   private String updatedByUserId;
 
   @UpdateTimestamp private OffsetDateTime updatedOn;
+
+  @ManyToOne(fetch = FetchType.LAZY)
```
vvartakMoJ: Can one Claim Amendment cause multiple calculated fee detail rows, or is it better (clearer) to make this a OneToOne relationship?

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/entity/RequestedByReference.java
```diff
@@ -0,0 +1,37 @@
...
+import java.time.OffsetDateTime;
+import java.util.UUID;
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+
+/** Entity representing a requested by reference. */
+@Getter
+@Setter
```
vvartakMoJ: As in the other reference entity, can we safely remove the @ Setter annotation, if this is an immutable entity?

claims-data/service/src/main/resources/db/migration/V40__seed_amendment_reference_data.sql
```diff
@@ -0,0 +1,36 @@
...
+-- We use a CTE (Common Table Expression) to temporarily store the generated UUIDs so we can reference them below
+WITH inserted_requesters AS (
+INSERT INTO claims.requested_by_reference (id, code, display_label, is_active, display_order)
+VALUES
+    (uuidv7(), 'PROVIDER', 'Provider', true, 10),
+    (uuidv7(), 'AUDITOR', 'Auditor', true, 20),
+    (uuidv7(), 'CONTRACT_MANAGEMENT', 'Contract Management', true, 30),
+    (uuidv7(), 'ASSURANCE', 'Assurance', true, 40)
+ON CONFLICT (code) DO UPDATE
+                          SET display_label = EXCLUDED.display_label, is_active = EXCLUDED.is_active
```
vvartakMoJ: Do we not want to allow updating of display_order, or should it also be included in the UPDATE statement?

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/entity/Claim.java
```diff
@@ -147,6 +151,14 @@ public class Claim {
   @Column(nullable = false)
   private Long version;
 
+  /**
+   * Convenience method to get the current/active fee calculation. Because of @OrderBy, the latest
+   * record is always at index 0.
+   */
+  public CalculatedFeeDetail getLatestCalculatedFee() {
```
vvartakMoJ, line 158: This method fetches all the rows even if we only want the top row. A named query or a repository method like findTopByClaimIdOrderByCreatedOnDesc might be preferable.

claims-data/service/src/integrationTest/java/uk/gov/justice/laa/dstew/payments/claimsdata/repository/ClaimAmendmentStorageIntegrationTest.java
```diff
@@ -0,0 +1,273 @@
...
+    CalculatedFeeDetail dynamicFeeUpdate =
+        amendClaimWithNewCalculation(targetClaim, typingErrorReason, BigDecimal.valueOf(999));
+
+    targetClaim.setCalculatedFeeDetails(new ArrayList<>());
+    targetClaim.getCalculatedFeeDetails().add(dynamicFeeUpdate);
+
+    claimRepository.flush();
+    Claim evaluatedClaim = claimRepository.findById(ClaimsDataTestUtil.CLAIM_1_ID).orElseThrow();
+
+    assertThat(evaluatedClaim.getCalculatedFeeDetails()).hasSize(1);
```
vvartakMoJ: If the fee details list size is only 1, this test is not really testing what it says (multiple calculations). The size here would be 2, if you remove line 117 above, i.e.

`targetClaim.setCalculatedFeeDetails(new ArrayList<>());`

vvartakMoJ (changes requested):
Left a few comments and suggestions.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/entity/ClaimAmendment.java
andrewJohnysJustice: These refer to the Jira ticket - where they are specified - but I'll remove them

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/entity/CalculatedFeeDetail.java
andrewJohnysJustice: I've made it one-to-many and we pick up the latest one - that way preserving history and it bullet proofs against data loss

andrewJohnysJustice: The method Claim.getLatestCalculatedFee() fetches the latest record and there's an integration test for that.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/entity/Claim.java
andrewJohnysJustice: I disagree - It gets the latest row - if it fetched a list then it would blow up as it's returning an object not a list. 
See test "shouldTrackMultipleCalculationsAndReturnLatestCFD() " in ClaimAmendmentStorageIntegrationTest

andrewJohnysJustice: Oh ! do you mean that it fetches all the rows and then filters them?

andrewJohnysJustice: So when the claim is fetched - it gets the list of CFD just for that claim. 
The get latest method returns the first item - but there's only likely to be a max of say a dozen at most, so I don't see a performance issue

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/entity/CalculatedFeeDetail.java
vvartakMoJ: I understand that one Claim can have multiple amendments (and hence multiple CFDs), this is specified in the Claim relationship on line 43 above.

My question is about the relationship between a Claim Amendment and CFD. I think each amendment will still have a single CFD, which is what this line is defining. Are you saying that one Claim Amendment can result in multiple CFDs?

vvartakMoJ (changes requested):

claims-data/service/src/main/resources/db/migration/V39__claim_amendment_storage_model.sql
```diff
@@ -53,7 +55,8 @@ CREATE INDEX ix_claim_amendment_claim_id
 CREATE INDEX ix_claim_amendment_created_on
     ON claims.claim_amendment (created_on DESC);
 
-CREATE INDEX ix_calculated_fee_detail_claim_amendment_id
+-- FIX: Changed to a UNIQUE index to optimize the One-to-One lookup path
```
vvartakMoJ: I think this unique index is redundant. Postgres always automatically creates a unique index for each unique constraint (viz. uq_calculated_fee_detail_claim_amendment).

vvartakMoJ (changes requested):

claims-data/service/src/main/resources/db/migration/V39__claim_amendment_storage_model.sql
andrewJohnysJustice: Removed

vvartakMoJ (dismissed):
Thanks for making the changes, LGTM!

claims-data/service/src/main/resources/db/migration/V39__claim_amendment_storage_model.sql
```diff
@@ -0,0 +1,59 @@
...
+);
+
+COMMENT ON TABLE claims.claim_amendment
+    IS 'Business record of successful claim amendments including requester and before/after state';
+
+
+-- -------------------------------------------------------------------
+-- 2. Alter Existing Tables
+-- -------------------------------------------------------------------
+ALTER TABLE claims.calculated_fee_detail
```
vvartakMoJ, line 25: This change WILL need to be coordinated with the GLAD team, as per the steps specified in the confluence page titled "Data Replication for Reporting Service".

andrewJohnysJustice:
Closing PR in order to tear the db down on the ephemeral branch

andrewJohnysJustice:
Re-opened

vvartakMoJ (approved):
LGTM!

PR #349, opened 2026-06-15
DSTEW-1763: Retrieve claim and build in-memory post-amendment state

vvartakMoJ:
## What
DSTEW-1763: Retrieve claim and build in-memory post-amendment state
[Link to story](https://dsdmoj.atlassian.net/browse/DSTEW-1763)

## Describe what you did and why.

Implements the in-memory retrieve-and-build step for the synchronous claim amendment flow (child of DSTEW-1593). Given a claim id and a sparse amendment payload, it loads the current claim aggregate, captures a before-state snapshot, and builds the proposed post-amendment state. No persistence and no external calls — a missing claim is returned as Optional.empty() for the parent flow to map to INVALID_CLAIM_NOT_FOUND.

## What's included

**DTOs**: ClaimAmendmentPayload (sparse, presence-aware via JsonNullable), ClaimStateSnapshot (immutable before/after snapshot), ClaimAmendmentState (before/payload/after aggregate), plus CalculatedFeeDetailSnapshot / AssessmentSnapshot.
**Mapper**: ClaimStateSnapshotMapper (MapStruct) assembles the before-state from the claim aggregate.
**Logic**: ClaimAmendmentStateBuilder applies tri-state semantics — omitted → unchanged, explicit null → cleared, value → applied; UCN/UFN are never recomputed from name/DOB/case-id changes.
**Service**: ClaimAmendmentStateService (@Transactional(readOnly = true)) orchestrates retrieval and build.
**Jackson**: registered JsonNullableModule on the primary ObjectMapper and added @Jacksonized so the omitted / explicit-null / value distinctions survive a serialise→deserialise round trip (basis for the future requestPayload/beforeState/diff JSONB columns).
**Tri-state semantics**: JsonNullable<T> preserves omitted (undefined) ≠ explicit null ≠ value end-to-end, with @JsonInclude(NON_ABSENT) dropping omitted fields and writing explicit nulls as "field": null.
## Tests
Unit tests for the mapper, builder, service, and payload (Jackson tri-state).
Integration tests (Testcontainers Postgres): end-to-end retrieve/build, not-found path, no DB side-effects, explicit-null clear, UCN/UFN not recomputed, latest-assessment selection, and real ObjectMapper tri-state serialisation/round-trip.
Centralised shared synthetic test data in AmendmentTestData; seeded values asserted by integration tests promoted to constants in AbstractIntegrationTest so seed data and assertions share a single source of truth.
## Notes
In-memory only; persistence of the amendment history JSONB columns is owned by a follow-up ticket.

## Checklist

### DIFF
--- a/claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/amendment/ClaimAmendmentStateBuilder.java
+++ b/claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/amendment/ClaimAmendmentStateBuilder.java
@@ -1,0 +1,257 @@
+package uk.gov.justice.laa.dstew.payments.claimsdata.service.amendment;
+
+import org.openapitools.jackson.nullable.JsonNullable;
+import org.springframework.stereotype.Component;
+import uk.gov.justice.laa.dstew.payments.claimsdata.dto.amendment.ClaimAmendmentPayload;
+import uk.gov.justice.laa.dstew.payments.claimsdata.dto.amendment.ClaimAmendmentState;
+import uk.gov.justice.laa.dstew.payments.claimsdata.dto.amendment.ClaimStateSnapshot;
+
+/**
+ * Builds the in-memory post-amendment state for a claim by applying a sparse amendment payload onto
+ * the current (before) state.
+ *
+ * <p>Sparse-payload semantics (see {@link ClaimAmendmentPayload}):
+ *
+ * <ul>
+ *   <li><b>omitted</b> field ({@code JsonNullable.undefined()}) &rarr; the stored value is
+ *       retained;
+ *   <li><b>explicit null</b> ({@code JsonNullable.of(null)}) &rarr; the value is cleared (kept as
+ *       {@code null} for later validation);
+ *   <li><b>value present</b> &rarr; the submitted value is applied.
+ * </ul>
+ *
+ * <p>UCN ({@code uniqueClientNumber}) and UFN ({@code uniqueFileNumber}) are treated like any other
+ * field: they change only when explicitly present in the payload. They are never recomputed from
+ * client forename, surname, date of birth or case id - applying each field independently guarantees
+ * this.
+ *
+ * <p>Read-only context (identity, lifecycle status, submission context, category of law, calculated
+ * fee detail and latest assessment) is carried over unchanged from the before-state.
+ *
+ * <p>This component performs no persistence and makes no external calls.
+ */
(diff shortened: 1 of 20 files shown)


claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/amendment/ClaimAmendmentStateService.java
```diff
@@ -0,0 +1,71 @@
...
+  private final ClaimAmendmentStateBuilder amendmentStateBuilder;
+
+  /**
+   * Retrieves the claim and builds the amendment state.
+   *
+   * @param claimId the claim identifier
+   * @param payload the sparse amendment payload as submitted
+   * @return the built amendment state, or {@link Optional#empty()} if the claim does not exist
+   */
+  @Transactional(readOnly = true)
```
BenJusticeDigital: Do we need this, the requesting method in the controller is already in a write transaction.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/amendment/ClaimAmendmentStateBuilder.java
```diff
@@ -0,0 +1,257 @@
...
+   * Applies the sparse payload onto the before-state and returns the proposed post-amendment state.
+   *
+   * @param beforeState the current stored claim state
+   * @param payload the sparse amendment payload as submitted
+   * @return the proposed post-amendment state
+   */
+  ClaimStateSnapshot buildPostAmendmentState(
+      ClaimStateSnapshot beforeState, ClaimAmendmentPayload payload) {
+
+    return beforeState.toBuilder()
```
BenJusticeDigital: This is very hard to follow may i make a suggestion to break the builder into logical chunks to make it clearer to read.

BenJusticeDigital (changes requested):

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/amendment/ClaimAmendmentStateBuilder.java
vvartakMoJ: Good idea. I split it up into logical groups and in the process was able to simplify the resolve method a bit too. Please take another look.

BenJusticeDigital (approved):
LGTM.

PR #353, opened 2026-06-18
[DSTEW-1751] - Added version to endpoint and wrote some tests. This m…

andrewJohnysJustice:
Updated current endpoint - added a mandatory version field. Tested the change.
Pact Tests currently fail, as Consumer will need to be updated first.

[Link to story](https://dsdmoj.atlassian.net/browse/DSTEW-1751)

## Checklist


andrewJohnysJustice:
Will bundle this into DSTEW-1752

PR #356, opened 2026-06-22
feat(DSTEW-1752): - Implemented 1658-B - Early Claim Version Gate

andrewJohnysJustice:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/DSTEW-1752)

Implemented 1658-B - Early Claim Version Gate so that any update to a claim must have a matching version to prevent stale claims from being updated

**Summary**
Integrated the new synchronous claim amendment validation orchestrator into the core ClaimService update flow. This includes enforcing a claim-status gate, migrating optimistic locking into a dedicated validation step, and updating the exception handling and API contracts to seamlessly support the new pipeline without breaking internal consumers.

**What's included**
Architecture & Integration

Orchestrator Wiring: Integrated ClaimAmendmentService and ClaimAmendmentStateService into ClaimService.updateClaim.

Claim-Status Gate: Added logic to ensure the synchronous amendment flow is only triggered when a claim is in a VALID status.

**Validation & Error Handling**

Version Validation Step: Extracted the hardcoded version check into ClaimVersionValidationStep, registering it as the first canonical step in the orchestrator pipeline.

Collected Exceptions: Created ClaimAmendmentValidationException to carry the full list of errors returned by the orchestrator.

Dynamic Status Routing: Added an @ExceptionHandler to inspect the error list and return a 409 Conflict if a version conflict is detected, or a standard 400 Bad Request for all other validation failures.

**Mapping & Contracts**

MapStruct JsonNullable: Added explicit OpenAPI JsonNullable overloads to ClaimMapper to bridge standard Java types with the generated amendment payload.

OpenAPI Relaxation: Removed version as a mandatory field in the OpenAPI spec to allow internal actors (like the event service) to perform status-only updates without triggering a 400 Bad Request.

**Testing**

Unit Tests: Refactored ClaimServiceTest to use @InjectMocks and updated assertions to expect the new orchestrated exception.

Integration Tests: Updated ClaimControllerIntegrationTest to seed claims in the VALID status, ensuring the new orchestrated flow is actually hit during test execution.

Pact Tests: Added global bypass mocks for the new state and orchestrator services to DataClaimsApiProviderTests to preserve existing consumer contracts.

## Checklist


claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/dto/amendment/ClaimAmendmentValidationCode.java
```diff
@@ -16,6 +16,10 @@
  */
 public enum ClaimAmendmentValidationCode {
 
+  /** The claim is voided and therefore cannot be amended. */
```
vvartakMoJ: Wrong comment description.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/exception/ClaimAmendmentValidationException.java
```diff
@@ -0,0 +1,21 @@
+package uk.gov.justice.laa.dstew.payments.claimsdata.exception;
+
+import java.util.List;
+import lombok.Getter;
+import uk.gov.justice.laa.dstew.payments.claimsdata.dto.amendment.ClaimAmendmentValidationError;
+
+/**
+ * Thrown when the synchronous claim amendment orchestrator detects validation failures. Carries the
+ * complete list of collected errors so the global exception handler can map them into a 400 Bad
```
vvartakMoJ: Slight correction: the errors include a ClaimAmendmentValidationCode which define their own http statuses (which is not always a 400 Bad Request, e.g. could be a 409 - Conflict, as in your case). We need to come up with a way to decide which one to surface if multiple codes exist in a given set of errors. My recommendation would be the HTTP code that belongs to the latest "fatal" error.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/exception/DataClaimsExceptionHandler.java
```diff
@@ -95,6 +96,43 @@ public ResponseEntity<ProblemDetail> handleGenericException(
...
+   *
+   * @param ex the exception containing the list of collected validation errors
+   * @return a {@link ResponseEntity} containing the list of errors, with an HTTP status of 409
+   *     (Conflict) or 400 (Bad Request) depending on the error contents
+   */
+  @ExceptionHandler(ClaimAmendmentValidationException.class)
+  public ResponseEntity<Object> handleClaimAmendmentValidationException(
+      ClaimAmendmentValidationException ex) {
+
+    // 1. Check if the error list contains the specific version conflict code
```
vvartakMoJ: I think this needs to be made more generic as we add more validation checks and have more than these two error codes (at a minimum, the decision should be based on error.getHttpStatus, not getCode).

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/amendment/validation/ClaimVersionValidationStep.java
```diff
@@ -0,0 +1,40 @@
...
+  /**
+   * Validates the requested amendment version against the current claim snapshot.
+   *
+   * @param state the in-memory amendment state containing both the before-state snapshot and the
+   *     requested patch payload
+   * @return a list containing an {@code INVALID_CLAIM_VERSION_CONFLICT} error if the versions do
+   *     not match, or an empty list if validation passes
+   */
+  @Override
+  public List<ClaimAmendmentValidationError> validate(ClaimAmendmentState state) {
```
vvartakMoJ, line 29: Please add unit tests for this class/method.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/dto/amendment/ClaimAmendmentPayload.java
```diff
@@ -57,6 +57,7 @@ public class ClaimAmendmentPayload {
   @Builder.Default private JsonNullable<String> procurementAreaCode = JsonNullable.undefined();
   @Builder.Default private JsonNullable<String> accessPointCode = JsonNullable.undefined();
   @Builder.Default private JsonNullable<String> deliveryLocation = JsonNullable.undefined();
+  @Builder.Default private JsonNullable<Long> version = JsonNullable.undefined();
```
vvartakMoJ, line 60: Logically this field does not belong here in the payload, it is not to be used to update the database record (and doesn't need the tri-state logic either). We use it for optimistic locking purposes and the version is eventually auto-incremented.

To avoid confusion (and exceptional treatment when calculating the diff to store in JSONB field later), I suggest that we don't add it here and instead add it to ClaimAmendmentState as "submittedVersion" and use it from there.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/exception/ClaimAmendmentValidationException.java
andrewJohnysJustice: In the DataClaimsExceptionHandler we could order the errors by status and throw the first one

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/amendment/validation/ClaimVersionValidationStep.java
andrewJohnysJustice: Added

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/ClaimService.java
```diff
@@ -182,7 +190,11 @@ public ClaimResponseV2 getClaimV2(UUID submissionId, UUID claimId) {
   @Transactional
   public void updateClaim(UUID submissionId, UUID claimId, ClaimPatch claimPatch) {
     Claim claim = requireClaim(submissionId, claimId);
+    if (claim.getStatus() == ClaimStatus.VALID) {
```
vvartakMoJ: I think this if statement should not be here. This is the job of a validation step (ClaimStatusValidationStep) which will check this and add an error to the list if required. Doing this here will silently ignore the user's invalid attempt to amend a claim.

andrewJohnysJustice: Well, yes - but we need away of steering away from existing code at the moment.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/ClaimService.java
```diff
@@ -217,6 +229,24 @@ public void updateClaim(UUID submissionId, UUID claimId, ClaimPatch claimPatch)
     }
   }
 
+  private void amendClaim(Claim claim, ClaimPatch claimPatch) {
+    ClaimAmendmentPayload payload = claimMapper.toAmendmentPayload(claimPatch);
+
+    ClaimAmendmentState state =
+        claimAmendmentStateService
+            .retrieveAmendmentState(claim.getId(), payload)
+            .orElseThrow(
```
vvartakMoJ: The caller (updateClaim) has already called requireClaim before this method is called, meaning it has already checked that the claim exists (or thrown an exception). I think retrieveAmendmentState should accept the whole claim (not claimId) from here and not bother with getting it again from the repository, which will also save a database call.

(I know this is not strictly part of this PR's remit but it is the first time this new method - retrieveAmendmentState - is being used and where this double DB access issue became visible).

vvartakMoJ (changes requested):
A few comments/suggestions.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/dto/amendment/ClaimAmendmentState.java
```diff
@@ -37,6 +37,8 @@ public class ClaimAmendmentState {
 
   private ClaimStateSnapshot postAmendmentState;
 
+  private Long submittedVersion;
+
```
BenJusticeDigital, line 53: I'm not fully clear on why we need this. The version passed is on the ClaimAmendmentPayload so why do we duplicate it here?

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/exception/DataClaimsExceptionHandler.java
BenJusticeDigital: This still needs to be resolved and all response entity's should be problem details like the other handlers. 
I think the logic should look for the first fatal if any and use the error and message as part of the problem detail.
If no fatal we should use a generic validation message and all errors shuold be under an issues node see the submisionValadationException for example. This is in main branch.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/ClaimService.java
```diff
@@ -183,6 +191,147 @@ public ClaimResponseV2 getClaimV2(UUID submissionId, UUID claimId) {
...
+
+    // Rule 2: Status is present, but there are other business fields attached
+    return hasAdditionalFieldUpdates(claimPatch);
+  }
+
+  /**
+   * Checks if the patch contains any fields outside of the standard status/fee update flow.
+   * Leverages short-circuit evaluation for maximum performance.
+   */
+  private boolean hasAdditionalFieldUpdates(ClaimPatch patch) {
```
BenJusticeDigital: For discussion use reflection and an allow list here. This seems like a maintenance headache and the down sides of reflection i.e. performance are negligable.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/ClaimService.java
```diff
@@ -183,6 +191,147 @@ public ClaimResponseV2 getClaimV2(UUID submissionId, UUID claimId) {
...
     Claim claim = requireClaim(submissionId, claimId);
 
+    if (isAnAmendment(claim, claimPatch)) {
+      amendClaim(claim, claimPatch);
+    } else {
+      updateClaimStatusAndFeeDetails(claim, claimPatch);
+    }
+  }
+
+  private boolean isAnAmendment(Claim claim, ClaimPatch claimPatch) {
```
BenJusticeDigital: Why are you passing a Claim here?

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/ClaimService.java
BenJusticeDigital: This still needs to be addressed, retrieveAmendmentState is still accepting the id not the claim. I also suspect  ClaimNotFoundException will never be actioned as for the optional to be empty it would have to have not found a claim and to get to this method it would need to have a claim.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/ClaimService.java
```diff
@@ -196,24 +345,31 @@ public void updateClaim(UUID submissionId, UUID claimId, ClaimPatch claimPatch)
...
-              });
+  private void amendClaim(Claim claim, ClaimPatch claimPatch) {
+    ClaimAmendmentPayload payload = claimMapper.toAmendmentPayload(claimPatch);
+
+    ClaimAmendmentState state =
+        claimAmendmentStateService
+            .retrieveAmendmentState(claim.getId(), payload, claimPatch.getVersion())
+            .orElseThrow(
+                () ->
+                    new ClaimNotFoundException(
```
BenJusticeDigital: All methods inside of the validation should add to issues and not throw their own exception. We should delegate to the ClaimAmendmentValidationException to deal with this.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/amendment/validation/ClaimVersionValidationStep.java
```diff
@@ -0,0 +1,43 @@
...
+    Long expectedVersion = state.getBeforeState().getVersion();
+    Long submittedVersion = state.getSubmittedVersion();
```
BenJusticeDigital: We should retrieve from the payload here.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/amendment/ClaimAmendmentService.java
```diff
@@ -55,7 +56,8 @@ public class ClaimAmendmentService {
       List.of(
           ClaimStatusValidationStep.class,
           AmendmentUserIdValidationStep.class,
-          AmendmentReferenceValidationStep.class);
+          AmendmentReferenceValidationStep.class,
+          ClaimVersionValidationStep.class);
```
BenJusticeDigital: The version seems a more important step than the others so i would promote this to 1st.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/ClaimService.java
```diff
@@ -183,6 +191,147 @@ public ClaimResponseV2 getClaimV2(UUID submissionId, UUID claimId) {
...
+        || patch.getSubmissionPeriod() != null
+        || patch.getIsAmended() != null
+        || patch.getHasAssessment() != null
+        || patch.getTotalWarnings() != null
+        || patch.getAmendmentRequestedBy() != null
+        || patch.getAmendmentUserId() != null
+        || patch.getAmendmentReasonCode() != null;
+  }
+
+  private void updateClaimStatusAndFeeDetails(Claim claim, ClaimPatch claimPatch) {
```
BenJusticeDigital, line 267: I would add a comment to this to explain the legacy decision that led to the need for this.

claims-data/service/src/test/java/uk/gov/justice/laa/dstew/payments/claimsdata/controller/ClaimControllerTest.java
```diff
@@ -172,7 +172,7 @@ void getClaimV2_returnsClaimDetails() throws Exception {
   void updateClaim_returnsNoContent() throws Exception {
     final UUID submissionId = Uuid7.timeBasedUuid();
     final UUID claimId = Uuid7.timeBasedUuid();
-    final String body = "{ \"status\": \"INVALID\" }";
+    final String body = "{ \"status\": \"INVALID\", \"version\": 1 }";
```
BenJusticeDigital: Is this still needed?

BenJusticeDigital (changes requested):

claims-data/service/src/integrationTest/java/uk/gov/justice/laa/dstew/payments/claimsdata/controller/ClaimControllerIntegrationTest.java
```diff
@@ -394,11 +393,14 @@ void shouldReturnAllClaimsForAGivenOfficeCode() throws Exception {
     // then: response body contains the expected number of claims
     String responseBody = result.getResponse().getContentAsString();
     var claimResultSet = OBJECT_MAPPER.readValue(responseBody, ClaimResultSet.class);
-    assertThat(claimResultSet.getTotalElements()).isEqualTo(3);
-    assertThat(claimResultSet.getContent()).hasSize(3);
+    assertThat(claimResultSet.getTotalElements()).isEqualTo(4);
```
vvartakMoJ: Please replace this "magic number" of 4 with a constant so that it only has to be changed once when new elements get added in future.

claims-data/service/src/integrationTest/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/amendment/ClaimAmendmentStateServiceIntegrationTest.java
```diff
@@ -80,7 +80,7 @@ void retrievesAmendmentState_happyPath_buildsBeforeAndPostState() {
             .build();
 
     Optional<ClaimAmendmentState> result =
-        amendmentStateService.retrieveAmendmentState(CLAIM_1_ID, payload);
+        amendmentStateService.retrieveAmendmentState(CLAIM_1_ID, payload, 1L);
```
vvartakMoJ: Literal 1L should also be replaced by a constant with a descriptive name such as SUBMITTED_VERSION_1 or something.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/dto/amendment/ClaimAmendmentState.java
vvartakMoJ: Yes, I think I floated it as an idea (because it bucks the trend of the tristate logic elsewhere) but the agreement at the end was to not have it as a separate parameter (i.e. this) but keep it in the payload (as we have the @ Version annotation in the repository which deals with the incrementing in a standard recognised way).

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/amendment/validation/ClaimVersionValidationStep.java
vvartakMoJ: And when we do, the test/IT should check that both explicit null and null by omission states are treated as null (i.e. the tri-state feature due to the use of JsonNullable).

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/amendment/ClaimAmendmentService.java
vvartakMoJ: For a definitive ordering see Abe's flowchart (https://dsdmoj.atlassian.net/wiki/spaces/WHCZA/pages/6144655993/Claim+Amendments+-+Validations+Flow) which shows this as the first step after checking that the claim ID exists.

claims-data/service/src/integrationTest/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/amendment/ClaimAmendmentStateServiceIntegrationTest.java
andrewJohnysJustice: It won't be there when I reverse the changes to the payload

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/ClaimService.java
andrewJohnysJustice: Added a Javadoc

claims-data/service/src/integrationTest/java/uk/gov/justice/laa/dstew/payments/claimsdata/controller/ClaimControllerIntegrationTest.java
andrewJohnysJustice: Added a constant

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/amendment/ClaimAmendmentService.java
andrewJohnysJustice: Rearranged

claims-data/service/src/test/java/uk/gov/justice/laa/dstew/payments/claimsdata/controller/ClaimControllerTest.java
andrewJohnysJustice: No

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/amendment/validation/ClaimVersionValidationStep.java
andrewJohnysJustice: That's how I did it originally

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/dto/amendment/ClaimAmendmentState.java
andrewJohnysJustice: We don't need it and I've removed it

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/exception/DataClaimsExceptionHandler.java
andrewJohnysJustice: I've reworked this.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/ClaimService.java
andrewJohnysJustice: .retrieveAmendmentState now passes the claim itself.

andrewJohnysJustice: We now do the requireClaim at the start of the service method and then we pass the claim around, so I have removed this. I think we can't add the claim not found as part of the amendment process itself, as we do it at the front door.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/exception/DataClaimsExceptionHandler.java
```diff
@@ -118,6 +122,56 @@ public ResponseEntity<ProblemDetail> handleGenericException(
...
+   */
+  @ExceptionHandler(ClaimAmendmentValidationException.class)
+  public ResponseEntity<Object> handleClaimAmendmentValidationException(
+      ClaimAmendmentValidationException ex) {
+
+    ClaimAmendmentValidationError error = ex.getErrors().getFirst();
+
+    if (error.isFatal()) {
+      Object responseBody = error.getMessage();
+      return ResponseEntity.status(error.getHttpStatus()).body(responseBody);
```
BenJusticeDigital: All errors should be returned as a ProblemDetail Object.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/exception/DataClaimsExceptionHandler.java
```diff
@@ -118,6 +122,56 @@ public ResponseEntity<ProblemDetail> handleGenericException(
...
+    }
+
+    Object responseBody = ex.getErrors();
+    return ResponseEntity.status(HttpStatus.BAD_REQUEST).body(responseBody);
```
BenJusticeDigital: All errors should be returned as a ProblemDetail Object. See handleSubmissionValidationException as an example of how to set one up with issues.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/exception/DataClaimsExceptionHandler.java
```diff
@@ -118,6 +122,56 @@ public ResponseEntity<ProblemDetail> handleGenericException(
...
+   * indicates to the client that their request was based on stale data and they must refresh their
+   * state before attempting the operation again.
+   *
+   * @param ex the optimistic locking exception thrown by the persistence layer
+   * @return a {@link ResponseEntity} containing a 409 status code and the exception message
+   */
+  @ExceptionHandler(ObjectOptimisticLockingFailureException.class)
+  public ResponseEntity<Object> handleDatabaseOptimisticLockingException(
+      ObjectOptimisticLockingFailureException ex) {
+    return ResponseEntity.status(HttpStatus.CONFLICT)
```
BenJusticeDigital: All errors should be returned as a ProblemDetail Object.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/amendment/validation/ClaimVersionValidationStep.java
```diff
@@ -0,0 +1,43 @@
...
+   *
+   * @param state the in-memory amendment state containing both the before-state snapshot and the
+   *     requested patch payload
+   * @return a list containing an {@code INVALID_CLAIM_VERSION_CONFLICT} error if the versions do
+   *     not match, or an empty list if validation passes
+   */
+  @Override
+  public List<ClaimAmendmentValidationError> validate(ClaimAmendmentState state) {
+    Long expectedVersion = state.getBeforeState().getVersion();
+    Long submittedVersion = state.getRequestPayload().getVersion().get();
```
BenJusticeDigital, line 36: As a precaution, we should consider checking that these objects are non-null before accessing nested values to prevent potential NPEs.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/ClaimService.java
```diff
@@ -183,6 +208,70 @@ public ClaimResponseV2 getClaimV2(UUID submissionId, UUID claimId) {
...
+    // Rule 2: Status is present, but there are other business fields attached
+    return hasAdditionalFieldUpdates(claimPatch);
+  }
+
+  /**
+   * Checks if the patch contains any fields outside of the standard status/fee update flow.
+   * Leverages short-circuit evaluation for maximum performance.
+   */
+  private boolean hasAdditionalFieldUpdates(ClaimPatch patch) {
+    if (patch == null) {
```
BenJusticeDigital, line 226: If this were true then the application would have NPE'd at line 220. May need to move this check to isAnAmendment and i would say throw an error.. no patch should never happen.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/ClaimService.java
```diff
@@ -80,6 +89,22 @@ public class ClaimService
...
+  private final ClaimAmendmentService claimAmendmentService;
+  private final ClaimAmendmentStateService claimAmendmentStateService;
+
+  private static final Set<String> IGNORED_FIELDS =
+      Set.of(
+          "serialVersionUID",
+          "id",
+          "status",
+          "version",
+          "feeCode",
```
BenJusticeDigital: There seem to be a lot more fields than i was expecting. Is there a reason for this? i was only expecting status, validationMessages and feeCalculationResponse.

BenJusticeDigital (changes requested):
A few comments + you seem to have a failing test

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/ClaimService.java
andrewJohnysJustice: That's true - I'll remove it

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/ClaimService.java
andrewJohnysJustice: I'll recheck them but there are a few fields used in the fee calculation part

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/exception/DataClaimsExceptionHandler.java
andrewJohnysJustice: The failing test was because of the feature switch - I've fixed it now

andrewJohnysJustice: "All errors should be returned as a ProblemDetail Object." - - ok - I get you - I'll build that in

andrewJohnysJustice: Fixed

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/amendment/validation/ClaimVersionValidationStep.java
andrewJohnysJustice: Added null checks and returns appropriate errors

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/ClaimService.java
andrewJohnysJustice: Based on a structural analysis of the updateClaimStatusAndFeeDetails method, here are the exact fields from ClaimPatch that are required for its execution, grouped by how they are utilized:

1. Explicitly Referenced Fields
These fields are directly checked or read via conditional logic within the method body:

validationMessages

Required for the initial null/empty safety checks and the .forEach() loop that writes validation logs to the database.

status

Extracted via claimPatch.getStatus() to pass to claimValidationService.ensureStatusIsNotVoid().

feeCalculationResponse

Required for the conditional block if (claimPatch.getFeeCalculationResponse() != null). It is passed into claimMapper.toCalculatedFeeDetail() to persist updated fee records.

createdByUserId

Extracted via claimPatch.getCreatedByUserId() inside the fee calculation block to auditing who initiated the calculation snapshot.

2. Implicitly / Structurally Required Fields
These fields are required under the hood by map constructors or framework contracts used by the method:

version

Even though it isn't explicitly written in this method, it is critically required for claimRepository.save(claim) to prevent lost updates via JPA optimistic locking when persisting the data state.

All Declared Field Properties (via claimMapper)

The method invokes claimMapper.updateSubmissionClaimFromPatch(claimPatch, claim). For the mapper to fully patch and update the target Claim entity state before persistence, it implicitly relies on any user-supplied domain value present in the payload (such as feeCode, quantities, rates, codes, etc.). Missing fields on the inbound object will leave matching properties unmutated on the targeted DB entity.

So I'll start with those fields and see what I can lose

andrewJohnysJustice: I know what's wrong. I am checking to see if there is a value in the Patch but I also need to check the Claim for the matching field to see if the values are equal and also ignore those !!

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/exception/DataClaimsExceptionHandler.java
```diff
@@ -118,6 +122,66 @@ public ResponseEntity<ProblemDetail> handleGenericException(
...
+    log.warn("ClaimAmendmentValidationException occurred with {} errors", ex.getErrors().size());
+    ClaimAmendmentValidationError primaryError = ex.getErrors().getFirst();
+
+    HttpStatus status = HttpStatus.BAD_REQUEST;
+    if (primaryError != null && primaryError.isFatal()) {
+      status = HttpStatus.resolve(primaryError.getHttpStatus().value());
+      if (status == null) {
+        status = HttpStatus.BAD_REQUEST;
+      }
+    }
```
vvartakMoJ: Nit: This block can be simplified to just:

status = primaryError.getHttpStatus();

because this is guaranteed to return a valid, non-null HTTP status enum value.

andrewJohnysJustice: I was being cautious, but I've simplified it now

vvartakMoJ (dismissed):
LGTM!

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/exception/ClaimAmendmentValidationException.java
```diff
@@ -0,0 +1,39 @@
...
+   *   <li>Errors are then sorted by their numeric HTTP status code values in descending order
+   *       (e.g., 500 before 400).
+   * </ol>
+   *
+   * @param errors the collected validation errors discovered during the amendment process
+   */
+  public ClaimAmendmentValidationException(List<ClaimAmendmentValidationError> errors) {
+    super("Claim amendment validation failed with " + errors.size() + " errors");
+
+    // Sort errors: fatal first (true -> false), then HTTP status code descending (500 -> 400)
```
BenJusticeDigital: General comments still reference the sortinvg

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/exception/DataClaimsExceptionHandler.java
```diff
@@ -118,6 +124,73 @@ public ResponseEntity<ProblemDetail> handleGenericException(
...
+        buildProblemDetailResponse(status, ex.getMessage(), ex.getClass(), request);
+
+    if (response.getBody() != null) {
+      response.getBody().setProperty("errors", ex.getErrors());
+    }
+
+    return response;
+  }
+
+  List<ClaimAmendmentValidationError> sortValidationErrorsByFatalAndStatus(
```
BenJusticeDigital, line 163: Is there a reason this method is package-private rather than explicitly marked private or public? Adding an access modifier could help make the intended visibility more obvious. I would also a JDoc to explain wht we order this way.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/ClaimService.java
```diff
@@ -80,6 +90,19 @@ public class ClaimService
...
+
+  private static final Set<String> IGNORED_FIELDS =
+      Set.of(
+          "id",
+          "submissionId",
+          "status",
+          "validationMessages",
+          "feeCalculationResponse",
+          "version",
+          "feeCode",
```
BenJusticeDigital: feeCode is a valid amendment. i may just want to update that values so i don't think this can be ignored.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/ClaimService.java
```diff
@@ -183,6 +206,94 @@ public ClaimResponseV2 getClaimV2(UUID submissionId, UUID claimId) {
...
+              Object claimValue = claimField.get(claim);
+
+              // If the values are completely identical, ignore it (no true business change)
+              if (Objects.equals(patchValue, claimValue)) {
+                return;
+              }
+            }
+
+            // If the matching entity field wasn't found, or the values are different:
+            log.info(
```
BenJusticeDigital: This could be very noisy, would this be better off as a debug?

BenJusticeDigital (changes requested):

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/exception/DataClaimsExceptionHandler.java
andrewJohnysJustice: I wrote a test for the method which is why I kept it package private - it tests the ordering of all the errors, but I can make it private and remove the test

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/ClaimService.java
andrewJohnysJustice: Yes, should be debug or removed, I added it to check what values were coming through

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/ClaimService.java
andrewJohnysJustice: Removed and fixed the Integration test

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/exception/ClaimAmendmentValidationException.java
andrewJohnysJustice: Updated comments - sorry - completely missed that one

BenJusticeDigital (dismissed):
LGTM

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/exception/ClaimAmendmentValidationException.java
```diff
@@ -0,0 +1,33 @@
+package uk.gov.justice.laa.dstew.payments.claimsdata.exception;
+
+import java.util.List;
+import lombok.Getter;
+import uk.gov.justice.laa.dstew.payments.claimsdata.dto.amendment.ClaimAmendmentValidationError;
+
+/**
+ * Exception thrown when one or more claim amendment validation steps fail.
+ *
+ * <p>This exception automatically sorts the underlying validation errors upon construction.
```
BenJusticeDigital: Comment still references sorting being done here.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/amendment/ClaimAmendmentService.java
```diff
@@ -51,29 +51,29 @@ public class ClaimAmendmentService {
...
+    PreparedAmendment prepared = preparationService.prepare(claim, payload);
 
     // Phase 2 - validate: run the ordered steps with no held transaction, so the inline external
     // PDA/FSP steps never hold a DB connection open.
     List<ClaimAmendmentValidationError> errors =
         validationService.validateAmendmentRequest(prepared.state());
     if (!errors.isEmpty()) {
       log.debug(
-          "Amendment for claim {} rejected with {} validation error(s)", claimId, errors.size());
+          "Amendment for claim {} rejected with {} validation error(s)", claim, errors.size());
```
BenJusticeDigital: should that not be claim.getId()

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/amendment/ClaimAmendmentService.java
```diff
@@ -51,29 +51,29 @@ public class ClaimAmendmentService {
...
       return ClaimAmendmentResult.rejected(errors);
     }
 
     // Phase 3 - commit: single atomic write transaction; reattaches the prepared claim.
     ClaimAmendment amendment = commitService.commit(prepared.claim(), prepared.state());
-    log.debug("Persisted amendment {} for claim {}", amendment.getId(), claimId);
+    log.debug("Persisted amendment {} for claim {}", amendment.getId(), claim);
```
BenJusticeDigital: same as previous comment

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/ClaimService.java
```diff
@@ -215,6 +324,20 @@ public void updateClaim(UUID submissionId, UUID claimId, ClaimPatch claimPatch)
                 validationMessageLogRepository.save(log);
               });
     }
+
+    // 3. Map the patch to the payload
+    uk.gov.justice.laa.dstew.payments.claimsdata.dto.amendment.ClaimAmendmentPayload payload =
```
BenJusticeDigital: Do we need the full class path?

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/ClaimService.java
```diff
@@ -196,15 +295,25 @@ public void updateClaim(UUID submissionId, UUID claimId, ClaimPatch claimPatch)
...
       calculatedFeeDetail.setCreatedByUserId(claimPatch.getCreatedByUserId());
       calculatedFeeDetailRepository.save(calculatedFeeDetail);
     }
+  }
+
+  private void amendClaim(Claim claim, ClaimPatch claimPatch) {
+    // 1. FAST FAIL: Optimistic Locking Guard
+    // Prevent heavy orchestration and external API calls if the client's version is already stale.
+    // A null version bypasses the check for internal system events.
+    if (claimPatch.getVersion() != null && !claimPatch.getVersion().equals(claim.getVersion())) {
```
BenJusticeDigital: This should not be needed as you have ClaimAmendmentValidationStep and should be picked up by if (result.errors() != null && !result.errors().isEmpty()) { with the correct and consistent error message.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/ClaimService.java
```diff
@@ -196,15 +295,25 @@ public void updateClaim(UUID submissionId, UUID claimId, ClaimPatch claimPatch)
...
+      throw new jakarta.persistence.OptimisticLockException("The record was modified concurrently");
+    }
 
+    // 2. Retain logic for saving incoming validation messages
```
BenJusticeDigital: This code is now duplicated and probably could be its own method. called from amendClaim and updateClaimStatusAndFeeDetails

claims-data/service/src/pactTest/java/uk/gov/justice/laa/dstew/payments/claimsdata/DataClaimsApiProviderTests.java
```diff
@@ -97,10 +99,23 @@ public class DataClaimsApiProviderTests extends AbstractProviderPactTests {
...
   @BeforeEach
   void setUp(PactVerificationContext context) {
     HttpTestTarget target = new HttpTestTarget("localhost", port);
     context.setTarget(target);
+
+    // 1. Capture the original state from the application context
+    originalAmendmentState = claimsApiProperties.getAmendments().isEnabled();
+  }
+
+  @org.junit.jupiter.api.AfterEach
```
BenJusticeDigital: Is the full path needed?

BenJusticeDigital (changes requested):
A couple of little NIT's round class paths, JavaDoc and code duplication. Also have you run the end to end on this to make sure that the Event service still can call the patch end point sucessfully.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/exception/ClaimAmendmentValidationException.java
andrewJohnysJustice: I'd removed that but then did a revert as it wasn't working...

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/ClaimService.java
andrewJohnysJustice: No - that was copilot - so I've removed it now

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/ClaimService.java
andrewJohnysJustice: I've removed it

claims-data/service/src/pactTest/java/uk/gov/justice/laa/dstew/payments/claimsdata/DataClaimsApiProviderTests.java
andrewJohnysJustice: No - copilot again - it seems to want to do it everywhere so I've now instructed it to stop

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/ClaimService.java
andrewJohnysJustice: Well yes, the rug was completely pulled yesterday

andrewJohnysJustice: Removed this as it's covered by Vid's code and my 3rd ticket will check for the functionality of this

BenJusticeDigital (approved):
LGTM, lets get this in and tested

vvartakMoJ (commented):
LGTM

PR #357, opened 2026-06-23
[DSTEW-1863] - Fix removing of valid chars from name

BenJusticeDigital:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/DSTEW-1863)

Preserve Unicode characters in client name fields (CSV/TXT bulk submissions)
### Problem
BulkSubmissionCsvConverter cleaned every field value with replaceAll("[^\\p{Print}]", ""). In Java \p{Print} is ASCII‑only, so this silently deleted any non‑ASCII character — accented letters and curly apostrophes — from imported data before validation. Valid names were corrupted rather than flagged, e.g. Siân → Sin, O’Brien → OBrien.
### Solution

- Clean per field instead of globally, keeping coded fields strict while letting names retain legitimate Unicode:
- NAME_FIELD_KEYS constant — single, easily‑amendable list of fields allowed to contain printable Unicode (CLIENT_FORENAME, CLIENT_SURNAME, CLIENT2_FORENAME, CLIENT2_SURNAME).
- Two small, documented helpers: stripNonPrintable (removes only non‑printable \p{C} — BOM/zero‑width/control) and stripToAscii (the original strict ASCII strip).
- cleanFieldValue(key, value) — applies the lenient strip to name fields, strict ASCII to everything else.
- getValues refactored into a small addCell helper (split first, then clean) so name Unicode is preserved while non‑name fields stay ASCII. Header cleaning unchanged (routed through stripToAscii).

### Tests
Added targeted unit tests for the three helpers (incl. BOM/zero‑width removal and null handling) and end‑to‑end convert() tests proving an accented/apostrophe surname survives while a non‑name field is still ASCII‑stripped. Suite: 30 → 41 tests, all green; Checkstyle/Spotless clean. The existing ASCII fixtures remain the structural regression net.

### Notes / follow‑ups (out of scope)
Downstream name validation pattern must allow both straight (') and curly (’) apostrophes, or names the converter now preserves could fail validation.

## Checklist


BenJusticeDigital:
Reworking as per new ticket

PR #358, opened 2026-06-23
DSTEW-1764: Synchronous claim amendment validation flow + claim-status gate

vvartakMoJ:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/DSTEW-1764)

## Summary
Introduces the synchronous claim amendment validation flow: an orchestrating service that runs an ordered, extensible list of validation steps over an already-retrieved amendment state, and delivers the first real gate — claim-status eligibility. The framework is deliberately built so the remaining steps (and the external calls / atomic save) drop in via their own tickets without reshaping this code.
## What's included
ClaimAmendmentService — runs each validation step in order, accumulating errors and short-circuiting on the first fatal error; applies the amendment only when every step passes (save itself is a follow-up).
ClaimAmendmentValidationStep — functional interface for a step: inspects the state, returns the errors it found (empty = pass). Some steps may make an external (PDA/FSP) call but are modelled as ordinary steps.
ClaimStatusValidationStep — the DSTEW-1764 gate: only VALID claims are amendable; a voided claim → INVALID_VOIDED_CLAIM_NOT_AMENDABLE, any other non-VALID status → INVALID_CLAIM_STATE_NOT_AMENDABLE (both fatal).
Error model — ClaimAmendmentValidationError (code + message + fatal flag) and new ClaimAmendmentErrorCode values. Contract: fatal stops the flow immediately; non-fatal errors accumulate so the caller sees every failure at once; any non-empty error set means nothing is saved.
Central ordering — AmendmentValidationConfig.STEP_ORDER is the single source of truth for the sequence; Spring-discovered step beans are sorted into it and exposed via the AmendmentValidationSteps wrapper. Fails fast at startup if the discovered steps don't match the declared order (forgotten or stray step).
## Design notes
Ordering lives in one place (the config), not scattered @ Order annotations, and is enforced at boot — so the documented sequence and the implemented steps can't silently drift.
Adding a step = write a @ Component implements ClaimAmendmentValidationStep + add it to STEP_ORDER.
Claim retrieval / not-found mapping remain the caller's responsibility (ClaimAmendmentStateService); this flow operates on the already-built before/after state.
## Out of scope / follow-ups (marked with TODO)
Remaining validation steps — claim-version contract (1751/1752), metadata (1765), changed-field (1766), amendability/assessed-claim (1767), fee-code (1768), duplicate (1769), outcome (1770), final version guard (1753/1754) — plus the PDA/FSP calls and the atomic save. These land in their own tickets.
## Testing
ClaimAmendmentServiceTest — ordering, fatal short-circuit, and that later steps don't run after a fatal error.
ClaimStatusValidationStepTest — eligible/ineligible outcomes and error codes per status.
AmendmentValidationConfigTest — STEP_ORDER integrity (no duplicates, matches the step components on the classpath) and the startup fail-fast.

## Checklist


claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/amendment/validation/ClaimStatusValidationStep.java
```diff
@@ -0,0 +1,48 @@
...
+              ClaimAmendmentErrorCode.INVALID_VOIDED_CLAIM_NOT_AMENDABLE,
+              VOIDED_CLAIM_MESSAGE,
+              true));
+    }
+
+    return List.of(
+        new ClaimAmendmentValidationError(
+            ClaimAmendmentErrorCode.INVALID_CLAIM_STATE_NOT_AMENDABLE,
+            CLAIM_STATE_NOT_AMENDABLE_MESSAGE.formatted(status, ClaimStatus.VALID),
+            true));
```
BenJusticeDigital: With the enhanced enum 
```
List.of(ClaimAmendmentErrorCode.INVALID_CLAIM_STATE_NOT_AMENDABLE.toValidationError(status, ClaimStatus.VALID)
```

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/dto/amendment/ClaimAmendmentValidationError.java
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * <p>Carries the stable {@link ClaimAmendmentErrorCode} and a human-readable message, so the parent
+ * flow can build the shared structured validation/error response.
+ */
+@Value
+public class ClaimAmendmentValidationError {
+
+  ClaimAmendmentErrorCode code;
+  String message;
+  boolean fatal;
```
BenJusticeDigital: Add Technical Message, and possibly make this a builder. It says immutable but with the enhanced class you may need to modify the technical message to add a lot of detail. message and fatal/severity can be imutable

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/config/AmendmentValidationConfig.java
```diff
@@ -0,0 +1,76 @@
...
+    Set<Class<?>> discovered =
+        discoveredSteps.stream().map(Object::getClass).collect(Collectors.toSet());
+    if (!discovered.equals(Set.copyOf(STEP_ORDER))) {
+      throw new IllegalStateException(
+          "Claim amendment validation steps do not match the declared order in "
+              + "AmendmentValidationConfig.STEP_ORDER. Declared="
+              + STEP_ORDER
+              + ", discovered="
+              + discovered);
+    }
```
BenJusticeDigital: Rather than make the system cross reference the list could you not just add a sortOrder method to the abstract class and check for duplicates and just order of that the default value being 9999. we could further add a disabled method. I think from what you have done we would need to delete the bean for it to not be actioned.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/dto/amendment/ClaimAmendmentErrorCode.java
```diff
@@ -0,0 +1,14 @@
+package uk.gov.justice.laa.dstew.payments.claimsdata.dto.amendment;
+
+/** Stable error codes produced by claim amendment validation. */
+public enum ClaimAmendmentErrorCode {
+
+  /** The claim is voided and therefore cannot be amended. */
+  INVALID_VOIDED_CLAIM_NOT_AMENDABLE,
```
BenJusticeDigital: To create a single point for error messages and state i would suggest that you enhance this enum by including all relevant data.
```
  private final String displayMessage;
  private final String technicalMessage;
  private final ValidationSeverity severity;

INVALID_VOIDED_CLAIM_NOT_AMENDABLE(
      "Claim status %s is not amendable; only claims with status %s can be amended.",
      "Technical Message...",
      ValidationSeverity.ERROR
  );

default ClaimAmendmentValidationError toValidationError(Object... params) {
    String code = ((Enum<?>) this).name();
    String message = String.format(getDisplayMessage(), params);
    return ClaimAmendmentValidationError.builder()
        .code(code)
        .message(message)
        .technicalMessage(getField())
        .severity(getSeverity())
        .build();
  }
```

ValidationSeverity = WARN, ERROR or FATAL

BenJusticeDigital (changes requested):

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/dto/amendment/ClaimAmendmentValidationError.java
vvartakMoJ: As discussed, I'll leave it to be added as part of a later PR which needs a technical message.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/dto/amendment/ClaimAmendmentErrorCode.java
vvartakMoJ: Done everything except the technicalMessage, to be added in a later PR.

claims-data/service/src/test/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/amendment/ClaimAmendmentServiceTest.java
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  @Test
+  @DisplayName("does not run later steps after a fatal error")
+  void stopsRunningLaterStepsAfterFatal() {
+    ClaimAmendmentValidationError fatal =
+        ClaimAmendmentValidationError.of(
+            ClaimAmendmentValidationCode.INVALID_VOIDED_CLAIM_NOT_AMENDABLE);
+    when(claimStatusValidationStep.validate(any())).thenReturn(List.of(fatal));
+    ClaimAmendmentValidationStep laterStep = mock(ClaimAmendmentValidationStep.class);
```
andrewJohnysJustice: We are mocking ClaimStatusValidationStep at class level and then using local mocks here - nothing wrong, but maybe do them all in the same way for consistency.

andrewJohnysJustice (commented):
I agree with Ben on the Enum front - just wants rounding out with those extra fields
Just a minor think with the test mocking for consistency

Happy to approve when you are ready

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/config/AmendmentValidationConfig.java
vvartakMoJ: As discussed, I have taken some of your suggestions and kept some of the original structure.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/amendment/ClaimAmendmentService.java
```diff
@@ -0,0 +1,127 @@
...
+   * Runs each validation step in order, stopping as soon as a fatal error is collected, and applies
+   * the amendment only when every step passes.
+   *
+   * @param state the in-memory amendment state produced by retrieval (DSTEW-1763)
+   * @return every validation error found; an empty list means validation passed and the amendment
+   *     was applied
+   */
+  public List<ClaimAmendmentValidationError> orchestrate(ClaimAmendmentState state) {
+    for (ClaimAmendmentValidationStep step : validationSteps) {
+      state.addErrors(step.validate(state));
```
BenJusticeDigital, line 87: We pass state to the step why don't we set the errors there rather than pass them back and then set them?  you could return state from the validator and do 
```
if(step.validate(state).containsFatal()){
   return state.getErrors();
}
```

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/amendment/validation/ClaimStatusValidationStep.java
```diff
@@ -0,0 +1,42 @@
...
+  public List<ClaimAmendmentValidationError> validate(ClaimAmendmentState state) {
+    ClaimStatus status = state.getBeforeState().getStatus();
+
+    if (status == ClaimStatus.VALID) {
+      return List.of();
+    }
+
+    if (status == ClaimStatus.VOID) {
+      return List.of(
+          ClaimAmendmentValidationError.of(
```
BenJusticeDigital, line 32: I think we could set the list directly to the state rather than pass it back.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/dto/amendment/ClaimAmendmentState.java
```diff
@@ -30,4 +36,26 @@ public class ClaimAmendmentState {
...
+
+  /** The validation errors collected so far as the amendment validation steps run. */
+  @Builder.Default private final List<ClaimAmendmentValidationError> errors = new ArrayList<>();
+
+  /**
+   * Adds the errors a validation step found to the running collection.
+   *
+   * @param newErrors the errors returned by a step; may be empty
+   */
+  public void addErrors(List<ClaimAmendmentValidationError> newErrors) {
```
BenJusticeDigital, line 48: I think a simple addError method is worth adding as if we update the state from the validator (see other comments) we can just add 1 at a time as weel as a list if needed.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/amendment/ClaimAmendmentService.java
vvartakMoJ: I've done it for clarity/separation of concerns:

1. No hidden side effects (i.e. the input parameter does not mutate in the call to validate)
2. When adding each step we do not have to remember to also add the errors to the state (the interface contract makes it explicit by virtue of requiring a return list of errors).

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/amendment/validation/ClaimStatusValidationStep.java
vvartakMoJ: See my response to the previous comment.

BenJusticeDigital (approved):
LGTM

siddharthshroff-MoJ:
LGTM

PR #361, opened 2026-06-25
[DSTEW-1757] - Method to determine if field effects FSP

BenJusticeDigital:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/DSTEW-1757)

Adds a single, auditable registry that answers one question: does a given claim field feed the Fee-Scheme-Platform (FSP) fee-scheme request for a given Area of Law? This is the lookup the amendment flow will use to decide whether a changed field requires the fee calculation to be (re)built.
### What changed

- New FeeSchemeRequestField enum (service/amendment/fee/), the single source of truth for the agreed claim to FSP request field mapping, including the nested bolt-on mappings. Each constant records:
- requestField - the FSP request field (target),
- claimField - the claim field it’s mapped from (source),
- areasOfLaw - the AreaOfLaw values the mapping applies to.
- Public lookup: static boolean mapsToFeeSchemeRequest(String field, AreaOfLaw areaOfLaw).
- New FeeSchemeRequestFieldTest - exhaustive coverage of the table.

### Behaviour / rules

- Most fields map for all areas of law; only the travel/waiting-cost fields are area-specific:
- travelWaitingCostsAmount → CRIME_LOWER and LEGAL_HELP,
- netWaitingCostsAmount → CRIME_LOWER only.
- A claim field can appear in multiple rows (different FSP targets per area); the lookup returns true if any matching row applies to the requested area.
- representationOrderDate is present in the registry but maps to no area (returns false), with a TODO - the FSP request field doesn’t exist yet. Kept visible so the mapping stays complete and auditable.
- Edge cases: unknown/null field = false; null AreaOfLaw = throws NullPointerException.

## Why an enum registry (not a util class)
Keeps every mapping in one reviewable place (mirroring the existing ClaimAmendmentValidationCode pattern), makes “all fields covered” auditable at a glance, and makes future changes (enabling representationOrderDate, fixing the bolt-on type) a one-line edit.
### Testing

- FeeSchemeRequestFieldTest - all 22 unconditional fields × 3 areas, both conditional fields, the unmapped representationOrderDate, unknown/null field, and the null-area throw. 34 cases, all passing.

### Notes for reviewers
- The registry isn’t wired into the validation pipeline yet, so you may see benign “unused” warnings until a caller (the fee-recalculation step) consumes it.
- mapsToFeeSchemeRequest matches on the claim-side field name. If a caller needs to match on the FSP request name instead, both values are already stored, so adding a mapsFromRequestField(...) variant is trivial.

## Checklist

### DIFF
--- a/claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/amendment/fee/FeeSchemeRequestField.java
+++ b/claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/amendment/fee/FeeSchemeRequestField.java
@@ -21,7 +21,7 @@ import uk.gov.justice.laa.dstew.payments.claimsdata.model.AreaOfLaw;
  * <ul>
  *   <li>{@link #getRequestField() requestField} - the FSP request field name (target);
  *   <li>{@link #getClaimField() claimField} - the claim field it is mapped from (source, and the
- *       value matched by {@link #mapsToFeeSchemeRequest(String, AreaOfLaw)});
+ *       value matched by {@link #impactsPricing(String, AreaOfLaw)});
  *   <li>{@link #getAreasOfLaw() areasOfLaw} - the areas of law for which the mapping applies. Most
  *       fields apply to every area; only the travel/waiting-cost fields are area-specific.
  * </ul>
@@ -88,7 +88,7 @@ public enum FeeSchemeRequestField {
    * @return {@code true} if the field maps to the fee-scheme request for that area of law
    * @throws NullPointerException if {@code areaOfLaw} is {@code null}
    */
-  public static boolean mapsToFeeSchemeRequest(String field, AreaOfLaw areaOfLaw) {
+  public static boolean impactsPricing(String field, AreaOfLaw areaOfLaw) {
     Objects.requireNonNull(areaOfLaw, "areaOfLaw must not be null");
     if (field == null) {
       return false;
--- a/claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/amendment/fee/FeeSchemeRequestField.java
+++ b/claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/amendment/fee/FeeSchemeRequestField.java
@@ -1,0 +1,108 @@
+package uk.gov.justice.laa.dstew.payments.claimsdata.service.amendment.fee;
+
+import java.util.Arrays;
+import java.util.EnumSet;
+import java.util.Objects;
+import java.util.Set;
+import lombok.Getter;
+import uk.gov.justice.laa.dstew.payments.claimsdata.model.AreaOfLaw;
+
+/**
+ * Central, auditable registry of the fee-scheme-platform (FSP) request fields and the claim field
+ * each is mapped from, together with the {@link AreaOfLaw areas of law} for which the mapping
+ * applies.
+ *
+ * <p>This is the single source of truth for "does this claim field feed the FSP fee-scheme
+ * request?". It is derived directly from the agreed mapping table (and its nested bolt-on table)
+ * and is deliberately exhaustive so the full mapping is reviewable in one place.
+ *
+ * <p>Each constant records:
+ *
+ * <ul>
+ *   <li>{@link #getRequestField() requestField} - the FSP request field name (target);
+ *   <li>{@link #getClaimField() claimField} - the claim field it is mapped from (source, and the
+ *       value matched by {@link #mapsToFeeSchemeRequest(String, AreaOfLaw)});
+ *   <li>{@link #getAreasOfLaw() areasOfLaw} - the areas of law for which the mapping applies. Most
+ *       fields apply to every area; only the travel/waiting-cost fields are area-specific.
+ * </ul>
+ *
+ * <p>A claim field may appear more than once when the same source feeds different FSP fields under
+ * different areas of law (e.g. {@code travelWaitingCostsAmount} &rarr; {@code netTravelCosts} for
+ * {@code CRIME_LOWER} and {@code travelAndWaitingCosts} for {@code LEGAL_HELP}).
+ */
+@Getter
+public enum FeeSchemeRequestField {
+  // ----- Main mapping table -----
+  FEE_CODE("feeCode", "feeCode", allAreas()),
+  CLAIM_ID("claimId", "id", allAreas()),
+  START_DATE("startDate", "caseStartDate", allAreas()),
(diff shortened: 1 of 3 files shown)


claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/amendment/fee/FeeSchemeRequestField.java
```diff
@@ -0,0 +1,108 @@
...
+   */
+  public static boolean mapsToFeeSchemeRequest(String field, AreaOfLaw areaOfLaw) {
+    Objects.requireNonNull(areaOfLaw, "areaOfLaw must not be null");
+    if (field == null) {
+      return false;
+    }
+    return Arrays.stream(values())
+        .filter(entry -> entry.claimField.equals(field))
+        .anyMatch(entry -> entry.areasOfLaw.contains(areaOfLaw));
+  }
```
naseberry, line 99: Nit: are we missng `@Nullable/@NonNull` on `mapsToFeeSchemeRequest` signature?

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/amendment/fee/FeeSchemeRequestField.java
```diff
@@ -0,0 +1,108 @@
...
+      "detentionTravelAndWaitingCosts", "detentionTravelWaitingCostsAmount", allAreas()),
+  CASE_CONCLUDED_DATE("caseConcludedDate", "caseConcludedDate", allAreas()),
+  NUMBER_OF_MEDIATION_SESSIONS("numberOfMediationSessions", "mediationSessionsCount", allAreas()),
+  JR_FORM_FILLING("jrFormFilling", "jrFormFillingAmount", allAreas()),
+  IMMIGRATION_PRIOR_AUTHORITY_NUMBER(
+      "immigrationPriorAuthorityNumber", "priorAuthorityReference", allAreas()),
+
+  // Row 21 - the FSP request field is not yet available, so this maps to no area of law and always
+  // returns false. Kept here so the mapping remains complete and auditable.
+  // TODO(DSTEW-xxxx): enable representationOrderDate once the FSP request field exists.
```
naseberry: Nit: Do we have a ticket for this, if not can we create a placeholder so this is not lost?

claims-data/service/src/test/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/amendment/fee/FeeSchemeRequestFieldTest.java
```diff
@@ -0,0 +1,110 @@
...
+  @DisplayName("travelWaitingCostsAmount maps for CRIME_LOWER and LEGAL_HELP")
+  void travelWaitingCostsMapsForCrimeLowerAndLegalHelp(AreaOfLaw areaOfLaw) {
+    assertThat(mapsToFeeSchemeRequest("travelWaitingCostsAmount", areaOfLaw)).isTrue();
+  }
+
+  @Test
+  @DisplayName("travelWaitingCostsAmount does not map for MEDIATION")
+  void travelWaitingCostsDoesNotMapForMediation() {
+    assertThat(mapsToFeeSchemeRequest("travelWaitingCostsAmount", AreaOfLaw.MEDIATION)).isFalse();
+  }
```
naseberry: Only checks `MEDIATION`. What if a new AreaOfLaw is added, would this test catch a regression?

Would it be ideal the follow the pattern below `netWaitingCostsMapsForCrimeLowerOnly` (lines 78–80) which manually lists `LEGAL_HELP` and `MEDIATION`?

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/amendment/fee/FeeSchemeRequestField.java
BenJusticeDigital: I reviewed this again as I was concerned about how it may have been missed, and I have identified that it was due to a misunderstanding on my part. This is the only value that exactly matches the claim-to-fee scheme request, so the mapping process handles it automatically. All other fields require explicit mapping, which is what initially caused the confusion.
Removed, enabled and added tests.

claims-data/service/src/test/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/amendment/fee/FeeSchemeRequestFieldTest.java
BenJusticeDigital: Done

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/amendment/fee/FeeSchemeRequestField.java
BenJusticeDigital: Nullness annotations (@Nullable/@NonNull) aren't a standard we've adopted  there are none on method signatures anywhere in the service module, and we have no static-analysis tooling (NullAway/Checker) wired into the build to enforce them, so they'd be purely decorative here.

Happy to revisit if we decide to adopt as a consistent standard but I don't think we should introduce a one-off here.

andrewJohnysJustice (dismissed):

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/amendment/fee/FeeSchemeRequestField.java
```diff
@@ -0,0 +1,104 @@
...
+
+  /**
+   * Whether the given claim field maps to the FSP fee-scheme request for the given area of law.
+   *
+   * @param field the claim field name (the "mapped from claim" value); may be {@code null}
+   * @param areaOfLaw the area of law to evaluate the mapping for; must not be {@code null}
+   * @return {@code true} if the field maps to the fee-scheme request for that area of law
+   * @throws NullPointerException if {@code areaOfLaw} is {@code null}
+   */
+  public static boolean mapsToFeeSchemeRequest(String field, AreaOfLaw areaOfLaw) {
```
vvartakMoJ: Another naming question!

Given this from the corresponding JIRA:
"This rule source is consumed by [DSTEW-1766](https://dsdmoj.atlassian.net/browse/DSTEW-1766) to populate ChangedFieldClassification.impacts_pricing."

is "impactsPricing" a better name for this method?

vvartakMoJ (commented):
Just a question for consideration.

vvartakMoJ (approved):
LGTM!

PR #388, opened 2026-07-15
feat(DSTEW-1768): Reject fee-code amendments that change Area of Law

BenJusticeDigital:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/DSTEW-XXX)

Adds a validation gate to the claim amendment flow that rejects a fee-code change when the new fee code belongs to a different Area of Law than the claim. Reuses the existing claims-validation-core package (no parallel validation path) - the Area of Law is taken from the validation result's resolvedData, so no extra lookup is introduced.
### Changes
- ClaimAmendmentValidationCode - new fatal error INVALID_FEE_CODE_AREA_OF_LAW_CHANGE (BAD_REQUEST), a terminal rejection.
- AmendmentExternalValidationStep - after the reusable validateClaim(...) call, when claim.feeCode is in the diff, compares the resolved feeSchemeAreaOfLaw against the claim's AreaOfLaw. The comparison is an exact match against the enum name form (e.g. LEGAL_HELP) that the Fee Scheme API returns; any non-exact value (including a differently-formatted one like LEGAL HELP) is treated as a different Area of Law and rejected. A null/absent Area of Law is left to the reusable validators' technical error (no duplicate gate rejection). Also aligned the step's usage to the claims-validation-core 1.1.0 API (ClaimValidationResult, Set<ClaimValidatorCode>).

### Tests
- AmendmentExternalValidationStepTest - migrated stale assertions to the 1.1.0 API and added gate unit tests: exact-match allowed, different/unknown/differently-formatted rejected, absent AoL fails via the technical error (not the gate), and fee-code-unchanged skips the gate.
- New ClaimAmendmentFeeCodeAreaOfLawIntegrationTest - end-to-end via the PATCH endpoint: same-AoL commits (write persisted), different-AoL rejected (nothing persisted), Fee Scheme 404/500/timeout controlled no-save paths, and fee-details caching behaviour (positive cache hit, distinct-code independence, per-fee-code positive-vs-missing-negative cache).
- MockServerIntegrationTest — added fee-details stub/verify helpers (stubFeeDetailsAreaOfLaw, stubFeeDetailsStatus, stubFeeDetailsWithDelay, verifyFeeDetailsCalled[ForFeeCode]); stub helpers now clear only expectations (not the request log).
- Fixture fix - responses/fee-scheme/get-fee-details-200.json now includes areaOfLaw: "LEGAL_HELP". The 1.1.0 bump made Area-of-Law resolution mandatory, so the fixture's missing field was causing all amendment-commit integration tests to fail; this repairs them.

### Notes
Exact match by design: the Fee Scheme returns the enum name form; a space/format variant is intentionally not a match.
Known upstream gap (flagged, not fixed here): claims-validation-core does not populate its short-lived negative cache for not-found/failed fee-code lookups (a 404 is thrown as an error and bypasses the switchIfEmpty → cacheNegative path), so repeated failing lookups re-hit the Fee Scheme API. The per-fee-code integration test documents this; a separate fix in the validator library is planned.

## Checklist


claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/amendment/validation/AmendmentExternalValidationStep.java
```diff
@@ -56,24 +61,10 @@
...
-        "CLAIM_SCHEDULE_REFERENCE",
-        "CLAIM_MANDATORY_FIELD",
-        "CLAIM_OUTCOME_CODE",
-        PDA_VALIDATION_STEP,
-        "CLAIM_UNIQUE_FILE_NUMBER"
-      };
+  private static final ClaimValidatorCode PDA_VALIDATION_STEP =
+      ClaimValidatorCode.CLAIM_CATEGORY_OF_LAW_VALIDATOR;
+
+  protected static final ClaimValidatorCode[] CLAIM_VALIDATOR_CODES = ClaimValidatorCode.values();
```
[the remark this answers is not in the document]
BenJusticeDigital: The expectation is that if validation were to broaden then we would want it to here as well so will not change.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/amendment/validation/AmendmentExternalValidationStep.java
```diff
@@ -102,6 +103,67 @@ public List<ClaimAmendmentValidationError> validate(ClaimAmendmentState state) {
...
+   * INVALID_FEE_CODE_AREA_OF_LAW_CHANGE} rejection.
+   *
+   * <p>The rule only applies when {@code claim.feeCode} is one of the changed diff fields. The Area
+   * of Law of the new fee code is read from the reusable validation package's {@link
+   * ResolvedClaimData#feeSchemeAreaOfLaw()} (populated during {@link
+   * ValidationService#validateClaim(Claim, Set)}); the claim's current Area of Law is taken from
+   * the before-state snapshot. The comparison is an <em>exact</em> match against the {@link
+   * Enum#name() name} form the Fee Scheme reports (e.g. {@code "LEGAL_HELP"}): any other value is
+   * treated as a different Area of Law. When the resolved Area of Law is absent, no rejection is
+   * raised here - that condition is surfaced as a fee-scheme technical issue by the reusable
```
vvartakMoJ, line 119: Is the technical issue surfaced as a fatal error or non-fatal? If it is non-fatal then perhaps it is something to consider (i.e. it won't stop the processing without having confirmed that the FSP derived AoL matches).

BenJusticeDigital: It won't stop processing but it also won't pass due to the technical error. This is an Abe question if we want to short circuit.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/amendment/validation/AmendmentExternalValidationStep.java
```diff
@@ -102,6 +103,67 @@ public List<ClaimAmendmentValidationError> validate(ClaimAmendmentState state) {
...
+        state.getBeforeState() == null ? null : state.getBeforeState().getAreaOfLaw();
+
+    // Nothing to compare against: a null/blank resolved area of law is surfaced by the reusable
+    // validators as a fee-scheme technical error, so we do not add a rejection here.
+    if (resolvedAreaOfLaw == null || resolvedAreaOfLaw.isBlank() || claimAreaOfLaw == null) {
+      return null;
+    }
+
+    // Exact match required. The Fee Scheme reports the area of law in the enum name form (e.g.
+    // "LEGAL_HELP"); any other value - including a differently-formatted one such as "LEGAL HELP" -
```
vvartakMoJ, line 152: Is it worth passing the AoL value through areaOfLawConverter.convert to resolve these kind of differences?

vvartakMoJ (approved):
LGTM, just one question, perhaps ignorable.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/amendment/validation/AmendmentExternalValidationStep.java
BenJusticeDigital: I went with you must match 100% to avoid any confusion. If this becomes a problem we can relax but my understanding is they should always match and have 1 ref per type.

PR #396, opened 2026-07-21
test: DSTEW-1935 integrate BDD suite into CI (phase 1 local mode)

MrRazaH3:
## What
Integrates the BDD (Cucumber) test suite into the CI pipeline as **Phase 1 – local mode**.

Jira: [DSTEW-1935](https://dsdmoj.atlassian.net/browse/DSTEW-1935)

## Why
The BDD suite lives in `claims-data/service` and exercises the bulk-submission and fee-scheme flows end-to-end, but until now it only ran on developer machines. This PR wires it into GitHub Actions so every PR and every push to `main` runs the full suite and blocks image publish on failure.

## How
Adds a sibling `bdd-test` job to both workflow files – it does **not** modify the shared MoJ reusable workflow.

- `.github/workflows/deploy-uat-preview.yml` – new job `needs: [build-and-test]`; added to `needs:` of `ecr-publish-image` and `deploy-uat-preview`.
- `.github/workflows/deploy-main.yml` – same job `needs: [release]`; added to `needs:` of `ecr-publish-image`.
- Runs `./gradlew :claims-data:service:bddTest` on `ubuntu-latest` (Docker preinstalled → Testcontainers Postgres works out of the box).
- `bdd.mode=local` (default) → PATCH-shortcut mock event-service. **No external network dependency, no secrets beyond `GITHUB_TOKEN`.**
- Uploads JUnit XML + HTML reports as `bdd-reports` artifact (`if: always()`, 14-day retention).

## Test evidence
Local run just before pushing:
```
./gradlew :claims-data:service:bddTest
BUILD SUCCESSFUL
76 tests, 0 failures, 0 errors
```

## Out of scope – Phase 2
Running the suite in `bdd.mode=uat` against the real event-service in UAT is deferred until:
- Runner topology decision (Peter)
- API-key header format from dev team
- Cleanup endpoints confirmed

## Files changed
- `.github/workflows/deploy-uat-preview.yml`
- `.github/workflows/deploy-main.yml`

2 files modified, 0 files created.

[DSTEW-1935]: https://dsdmoj.atlassian.net/browse/DSTEW-1935?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/bdd-test/action.yml
+++ b/.github/actions/bdd-test/action.yml
@@ -1,0 +1,53 @@
+name: "BDD test"
+description: "Run the claims-data BDD (Cucumber) suite in local mode and upload reports."
+
+inputs:
+  java-version:
+    description: "JDK version to install"
+    required: false
+    default: "25"
+  java-distribution:
+    description: "JDK distribution"
+    required: false
+    default: "corretto"
+  github-actor:
+    description: "Actor used to authenticate against GitHub Packages"
+    required: true
+  github-token:
+    description: "Token used to authenticate against GitHub Packages (needs packages:read)"
+    required: true
+  artifact-name:
+    description: "Name of the uploaded reports artifact"
+    required: false
+    default: "bdd-reports"
+
+runs:
+  using: "composite"
+  steps:
+    - name: Set up JDK
+      uses: actions/setup-java@0f481fcb613427c0f801b606911222b5b6f3083a # v5.5.0
+      with:
+        java-version: ${{ inputs.java-version }}
+        distribution: ${{ inputs.java-distribution }}
+
+    - name: Set up Gradle
+      uses: gradle/actions/setup-gradle@3f131e8634966bd73d06cc69884922b02e6faf92 # v6.0.0
+
+    - name: Run BDD test suite (local mode)
+      shell: bash
+      env:
+        GITHUB_ACTOR: ${{ inputs.github-actor }}
+        GITHUB_TOKEN: ${{ inputs.github-token }}
+      run: ./gradlew :claims-data:service:bddTest --no-daemon
+
+    - name: Upload BDD reports
+      if: always()
+      uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+      with:
+        name: ${{ inputs.artifact-name }}
+        path: |
+          claims-data/service/build/reports/tests/bddTest/**
+          claims-data/service/build/test-results/bddTest/**
+        if-no-files-found: warn
+        retention-days: 14
+
--- a/.github/actions/bdd-test/action.yml
+++ b/.github/actions/bdd-test/action.yml
@@ -10,12 +10,6 @@ inputs:
     description: "JDK distribution"
     required: false
     default: "corretto"
-  github-actor:
-    description: "Actor used to authenticate against GitHub Packages"
-    required: true
-  github-token:
-    description: "Token used to authenticate against GitHub Packages (needs packages:read)"
-    required: true
   artifact-name:
     description: "Name of the uploaded reports artifact"
     required: false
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy-uat-preview.yml
```diff
@@ -102,6 +102,43 @@ jobs:
...
+        if: always()
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: bdd-reports
+          path: |
+            claims-data/service/build/reports/tests/bddTest/**
+            claims-data/service/build/test-results/bddTest/**
+          if-no-files-found: warn
+          retention-days: 14
+
```
vvartakMoJ, line 121: This code is shared between the two workflows and can/should be extracted into a shared action (.github/actions/bdd-test)

vvartakMoJ (changes requested):
Looks good, it would be better if we put the shared code in a separate action and reuse it from the two workflows.

.github/workflows/deploy-uat-preview.yml
MrRazaH3: Thanks Vid — good call. Extracted the job body into a composite action at `.github/actions/bdd-test/action.yml` and both workflows now just call `uses: ./.github/actions/bdd-test` with `java-version` / `java-distribution` / `github-actor` / `github-token` as inputs. The `needs:` wiring in each workflow is unchanged. Pushed as 09ecc79e.

MrRazaH3:
Pushed **09ecc79e** — extracted the duplicated `bdd-test` job body into a composite action at `.github/actions/bdd-test/action.yml`.

**What moved into the action**
- `actions/setup-java` (JDK)
- `gradle/actions/setup-gradle`
- `./gradlew :claims-data:service:bddTest` run (with `GITHUB_ACTOR` / `GITHUB_TOKEN` for GitHub Packages auth)
- `actions/upload-artifact` for the HTML + JUnit-XML reports (`if: always()`, 14-day retention)

**What stayed in each workflow**
- The `bdd-test` job shell itself (runner, `needs:`, `permissions`, `timeout-minutes`, checkout)
- The `needs: [ ..., bdd-test ]` wiring on `ecr-publish-image` / `deploy-uat-preview`

**Inputs on the action**: `java-version` (default `25`), `java-distribution` (default `corretto`), `github-actor`, `github-token`, `artifact-name` (default `bdd-reports`).

Net diff: `+63 / -46` across 3 files (1 new, 2 slimmer). YAML parses cleanly and local `bddTest` still passes (76/0).

.github/actions/bdd-test/action.yml
```diff
@@ -0,0 +1,53 @@
...
+        distribution: ${{ inputs.java-distribution }}
+
+    - name: Set up Gradle
+      uses: gradle/actions/setup-gradle@3f131e8634966bd73d06cc69884922b02e6faf92 # v6.0.0
+
+    - name: Run BDD test suite (local mode)
+      shell: bash
+      env:
+        GITHUB_ACTOR: ${{ inputs.github-actor }}
+        GITHUB_TOKEN: ${{ inputs.github-token }}
```
vvartakMoJ: I think these two values are directly available to the composite actions, i.e. you should be able to replace these lines with this:

        GITHUB_ACTOR: ${{ github.actor }}
        GITHUB_TOKEN: ${{ github.token }}

This will allow you to remove the two input parameters from this file as well as the two calling workflows.

vvartakMoJ (commented):
Sorry, one more request.

.github/actions/bdd-test/action.yml
MrRazaH3: Nice — you're right, that's cleaner. Dropped the `github-actor` / `github-token` inputs from the action and both workflow `with:` blocks, and the step now uses `${{ github.actor }}` / `${{ github.token }}` directly. The action's public inputs are now just `java-version`, `java-distribution`, and `artifact-name`. Pushed as e155eb1a.

vvartakMoJ (approved):
LGTM!

PR #397, opened 2026-07-21
feat(DSTEW-1948): Sort claims based on derived status

BenJusticeDigital:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/DSTEW-1948)

The raw claim_status (READY_TO_PROCESS, VALID, INVALID, VOID) only describes a claim's validation lifecycle. The business needs a single, friendlier status that also reflects assessment/amendment state. This PR introduces DerivedClaimStatus - a read-only value derived from claim_status, has_assessment and is_amended - exposes it on the v2 claims search response, and makes it a sortable field. The raw status is unchanged and v1 is unaffected.
Derivation (first match wins, fixed precedence - not chronological): VOID → VOIDED · INVALID → INVALID · READY_TO_PROCESS → READY_TO_PROCESS · has_assessment → ASSESSED · is_amended → AMENDED · else (VALID) → ACCEPTED. Null flags are treated as false.
Canonical order (asc): ACCEPTED, AMENDED, ASSESSED, VOIDED, INVALID, READY_TO_PROCESS (desc = reverse).
### Key changes

- OpenAPI spec - new derived_claim_status enum (declaration order is the canonical ordering) and field on claim_response_v2 only; documented the v2 sort param.
- DerivedClaimStatusResolver (new) - the single source of truth for the derivation, consumed by response mapping.
- ClaimMapper -populates derivedClaimStatus; raw status left untouched.
- ClaimSortField / ClaimSpecification / ClaimService - added derived_claim_status as a computed sort via a SQL CASE whose ordinals come from DerivedClaimStatus.ordinal() (so precedence/order is defined in exactly one place). Sorting runs in the DB across the full result set.
- Deterministic ordering -appended a claim.id ASC (UUIDv7) tie-break to all v2 claims sorts (plain + the existing total_warnings/submission_period computed sorts) for stable pagination.
- Bug fix - made the pageable handling unpaged-safe; a sort-without-page request previously threw and returned 500.

### Single source of truth

- Derivation logic lives only in DerivedClaimStatusResolver; ordering ordinals come only from the DerivedClaimStatus enum. A parity/integration test guarantees the SQL ordering matches the resolver so the two can't drift.

### Error handling

- Unsupported sort keys still return the existing 400 Problem Details response.
- Multi-sort caveat (documented)
- Plain columns can be combined freely (+ id tie-break). Computed fields behave as single-field sorts and shouldn't be combined with other sort fields - matching the pre-existing total_warnings/submission_period behaviour. Docs outline how full computed multi-field sorting could be added later.

### Tests

- Unit: resolver truth table (16 cases) + null/canonical-order; mapper (derived populated, raw status preserved); service (unsupported key → 400, tie-break applied, computed sort stripped, unpaged-safe).
- Integration: DB-level asc/desc ordering across all six statuses, id ASC pagination stability, 400 for unknown keys.
- Full suite green (1076 unit + integration), checkstyle and spotless clean.

### Docs
docs/derived-claim-status.md - purpose, rationale, algorithm, truth table, enum, sort behaviour, examples, relationship to raw claim_status, implementation notes, test expectations, and the multi-sort caveat.
### Notes for reviewers

- The id tie-break was applied to all v2 sorts (not just the new one) for consistency -this makes previously non-deterministic sorts stable.
- The unpaged 500 fix is pre-existing behaviour that the new sortable field made more likely to be hit; covered by a regression test.

## Checklist


andrewJohnysJustice:
Failed on my build with spotlessApply

andrewJohnysJustice (dismissed):

BenJusticeDigital:
> Failed on my build with spotlessApply

Thanks @andrewJohnysJustice That should now be resolved for a new approval as the fix dismissed your original one.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/ClaimService.java
```diff
@@ -589,7 +606,41 @@ private Pageable removeCustomSortFromPageable(Pageable pageable, String customPr
 
     Sort newSort = remainingOrders.isEmpty() ? Sort.unsorted() : Sort.by(remainingOrders);
 
+    if (pageable.isUnpaged()) {
+      return newSort.isSorted() ? Pageable.unpaged(newSort) : Pageable.unpaged();
+    }
+
     return org.springframework.data.domain.PageRequest.of(
```
andrewJohnysJustice: Do we need the package here?
Might be one of mine, tbh
It wasn't me or thee

andrewJohnysJustice (dismissed):

andrewJohnysJustice (approved):

vvartakMoJ (approved):
LGTM!

PR #406, opened 2026-07-28
feat(DSTEW-1644): submission totals recalculation

andrewJohnysJustice:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/DSTEW-1644)

This PR updates the submission read calculations to ensure submission totals use only the latest calculated_fee_detail row for each claim. This is a critical read-side correctness fix that paves the way for the one-to-many calculated fee detail storage model, preventing the double-counting of original and amended fee rows.

Key Changes
Single Submission Recalculation: Refactored getCalculatedTotalAmount in SubmissionRepository from a standard JPQL SUM to a native SQL query. It now uses a ROW_NUMBER() OVER (PARTITION BY cfd.claim_id ORDER BY cfd.created_on DESC, cfd.id DESC) window function to strictly filter out historical amendment rows before calculating the total.

Bulk Submission Recalculation: Applied the exact same window function logic to the plural getCalculatedTotalAmounts method, ensuring dashboard and list views accurately roll up the partitioned totals grouped by submission_id via a native query.

Validation Pipeline Fixes: Updated the ClaimStateSnapshot test fixtures in ClaimAmendmentValidationServiceIntegrationTest to inject a baseline CalculatedFeeDetailSnapshot. This aligns the integration tests with the strict rejection logic recently added to the AmendmentFspValidationStep.

Strict Test Builders: Expanded the test data builders inside SubmissionRepositoryIntegrationTest to satisfy mandatory database constraints (provider_user_id, created_by_user_id, submission_period, area_of_law, line_number, matter_type_code).

Testing Performed
Added a new nested integration test suite (SubmissionTotalsRecalculation) within SubmissionRepositoryIntegrationTest that hits the live test database to prove the Sum Calculation and native queries execute correctly. Tests include:

Single calculated-fee row per claim sums normally (Regression).

Multiple calculated-fee rows per claim sum only the latest row based on created_on.

Tie-break scenario verifying id fallback when multiple rows share the exact same created_on timestamp.

Bulk Query Validation: Verifies that multiple submissions passed to the IN clause securely partition and sum the latest fees per submission without cross-pollinating totals.

## Checklist

### DIFF
--- a/claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/repository/SubmissionRepository.java
+++ b/claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/repository/SubmissionRepository.java
@@ -41,7 +41,7 @@ public interface SubmissionRepository
   @Query(
       value =
           """
-          SELECT COALESCE(SUM(latest_fees.total_amount), 0)
+          SELECT SUM(latest_fees.total_amount)
           FROM (
             SELECT cfd.total_amount,
                    ROW_NUMBER() OVER (PARTITION BY cfd.claim_id ORDER BY cfd.created_on DESC, cfd.id DESC) as rn
--- a/claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/repository/SubmissionRepository.java
+++ b/claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/repository/SubmissionRepository.java
@@ -57,9 +57,9 @@ public interface SubmissionRepository
   /**
    * Returns calculated total amounts for the given submissions.
    *
-   * <p>For each submission ID provided, this query returns the sum of {@code
-   * calculatedTotalInclVat} from the latest cfd record for each claim belonging to that submission.
-   * Results are grouped by submission ID.
+   * <p>For each submission ID provided, this query returns the sum of {@code totalAmount} from the
+   * latest cfd record for each claim belonging to that submission. Results are grouped by
+   * submission ID.
    *
    * <p>Submissions with no cfd records are not included in the returned list.
    *
--- a/claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/repository/SubmissionRepository.java
+++ b/claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/repository/SubmissionRepository.java
@@ -33,6 +33,15 @@ public interface SubmissionRepository
     /**
      * Returns the Calculated total amount for the submission.
      *
+     * <p>This aggregation guarantees the following behavior:
+     *
+     * <ul>
+     *   <li>DSTEW-1538 - Must use scaleNullable
+     *   <li>Returns {@code null} if there are no CFD rows associated with the submission.
+     *   <li>Returns {@code 0} (Zero) if CFD rows exist but their combined sum is exactly zero.
+     *   <li>Returns the exact summed total of the latest CFD rows for all other scenarios.
+     * </ul>
+     *
      * @return the summed Calculated total amount
      */
     BigDecimal getTotal();
@@ -70,7 +79,7 @@ public interface SubmissionRepository
       value =
           """
           SELECT latest_fees.submission_id AS submissionId,
-                 COALESCE(SUM(latest_fees.total_amount), 0) AS total
+                 SUM(latest_fees.total_amount) AS total
           FROM (
             SELECT c.submission_id,
                    cfd.total_amount,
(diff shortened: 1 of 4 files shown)


siddharthshroff-MoJ:
Can we add more assertions to cover acceptance criteria 5(response shapes are unchanged) ?

siddharthshroff-MoJ (dismissed):
LGTM

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/repository/SubmissionRepository.java
```diff
@@ -39,34 +39,50 @@ interface CalculatedTotalAmountProjection {
...
+          FROM (
+            SELECT cfd.total_amount,
+                   ROW_NUMBER() OVER (PARTITION BY cfd.claim_id ORDER BY cfd.created_on DESC, cfd.id DESC) as rn
+            FROM claims.calculated_fee_detail cfd
+            INNER JOIN claims.claim c ON c.id = cfd.claim_id
+            WHERE c.submission_id = :submissionId
+          ) latest_fees
+          WHERE latest_fees.rn = 1
+          """,
+      nativeQuery = true)
```
BenJusticeDigital, line 63: Just a thought on this query: SUM(latest_fees.total_amount) will return NULL if there are no rows after WHERE latest_fees.rn = 1, and it will also return NULL if all matching total_amount values are NULL. Individual NULL values are ignored by SUM(), but an empty result set produces NULL.

Looking at the entity/schema, total_amount appears to be nullable (BigDecimal totalAmount with no @NotNull/nullable = false, and the column was added as NUMERIC without NOT NULL), so both scenarios seem possible.

Could we add tests for:

No matching claims/fees for the submission (no rows returned)
Matching rows where total_amount is NULL
A mix of populated and NULL total_amount values

We should also make an explicit decision on the expected business behaviour when no amount is available: should the total be 0 (using COALESCE(SUM(...), 0)) or should it remain NULL to indicate "no value calculated"? That will help avoid unexpected failures and make the intent clear.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/repository/SubmissionRepository.java
```diff
@@ -39,34 +39,50 @@ interface CalculatedTotalAmountProjection {
...
+                   cfd.total_amount,
+                   ROW_NUMBER() OVER (PARTITION BY cfd.claim_id ORDER BY cfd.created_on DESC, cfd.id DESC) as rn
+            FROM claims.calculated_fee_detail cfd
+            INNER JOIN claims.claim c ON c.id = cfd.claim_id
+            WHERE c.submission_id IN (:submissionIds)
+          ) latest_fees
+          WHERE latest_fees.rn = 1
+          GROUP BY latest_fees.submission_id
+          """,
+      nativeQuery = true)
```
BenJusticeDigital, line 94: See previous comment as that applies here as well.

BenJusticeDigital (changes requested):

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/repository/SubmissionRepository.java
andrewJohnysJustice: That's what Abe wanted. If there are no rows it should return NULL. 
If there are rows but totalling ZERO then it should return zero
Anything else returns a value

BenJusticeDigital: Ok lets document that decision with the SQL so future me does not forget and add tests for no rows, all null rows and a mixture. That way if someone messes with this it will break the tests.

andrewJohnysJustice: Sure I can do that - the fundamental way that the SQL works hasn't changed - the only thing that should have changed is that we only regard the latest CFD per claim - otherwise it should still behave exactly the same as it did previously, and we have existing tests to prove that, plus new tests which failed before I changed the SQL as it was picking up all the CFDs

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/repository/SubmissionRepository.java
andrewJohnysJustice: Submissions with no CFDs are not included in the response List

vvartakMoJ (dismissed):
LGTM

BenJusticeDigital (dismissed):

BenJusticeDigital (approved):

PR #410, opened 2026-07-29
feat(DSTEW-2058): Enforce unique claim line number within a submission (409 on duplicate)

BenJusticeDigital:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/DSTEW-2058)

There was previously no uniqueness enforcement on a claim's line_number within a submission, so duplicate rows were silently persisted to the database. This PR enforces uniqueness of (submission_id, line_number) and returns a 409 Conflict when a duplicate is submitted, without breaking environments that already contain historical duplicates (business rule: we never amend or delete historical data).
### Three-layer enforcement

- DB (authoritative): Flyway migration V45 adds a partial unique index uq_claim_submission_line_number on claim (submission_id, line_number), gated by a configurable Flyway placeholder cutoff (claim_line_number_uniqueness_cutoff, defaults to epoch = enforce all rows). Using a partial index lets environments with existing duplicates adopt the constraint by setting a cutoff, rather than failing the index build.
- Application pre-check: ClaimService.createClaim calls existsBySubmissionIdAndLineNumber and throws a typed DuplicateClaimException (409) for a clean, fail-fast response that also covers the old-vs-new gap.
- Handler backstop: DataClaimsExceptionHandler maps DataIntegrityViolationException for this specific index to 409 (RFC 9457 problem detail) to catch the residual concurrency race.

### Also included

- OpenAPI spec: documented 409 on createClaim.
- Pact provider state for the duplicate-create scenario.
- Unit, integration, and constraint tests; release note with pre-deploy duplicate-count query and cutoff guidance.

### Caveats (documented in code)

- Old-vs-new: a partial index can't catch a new row duplicating a grandfathered pre-cutoff row (currently unreachable as claims are only added to new submissions).
- A minimal TOCTOU race remains between the pre-check and insert, closed by the DB index for post-cutoff rows.

### Notes for reviewers
Environments with existing duplicates must set CLAIM_LINE_NUMBER_UNIQUENESS_CUTOFF before deploy (see release note).

## Checklist


claims-data/service/src/main/resources/db/migration/V45__add_unique_constraint_claim_submission_line_number.sql
```diff
@@ -0,0 +1,26 @@
...
+-- this gap defensively (with a minimal residual race) should that rule ever change.
+--
+-- The index is named uq_claim_submission_line_number so that unique-violation errors report that
+-- name, which DataClaimsExceptionHandler uses to map the violation to HTTP 409 as a backstop.
+--
+-- NOTE: for a very large table in an environment with existing data, consider building this index
+-- with CREATE UNIQUE INDEX CONCURRENTLY in a non-transactional migration to avoid a write lock.
+CREATE UNIQUE INDEX uq_claim_submission_line_number
+    ON claim (submission_id, line_number)
+    WHERE created_on > TIMESTAMPTZ '${claim_line_number_uniqueness_cutoff}';
```
naseberry, line 25: Is this something we need to consider given we have have millions of claims?

BenJusticeDigital: I would defer to @vvartakMoJ but my thoughts are this is done when we have no traffic (out of hours) then it should not be an issue.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/ClaimService.java
```diff
@@ -123,6 +124,29 @@ public Supplier<SubmissionNotFoundException> entityNotFoundSupplier(String messa
...
+    // RACE: this check is not atomic with the insert below, so a small TOCTOU window remains if two
+    // requests create the same (submission_id, line_number) concurrently. The DB unique index
+    // closes that window for the common (post-cutoff) case; the residual race only affects the
+    // old-vs-new scenario above and is considered minimal/acceptable.
+    Integer lineNumber = claimPost.getLineNumber();
+    if (lineNumber != null
+        && claimRepository.existsBySubmissionIdAndLineNumber(submissionId, lineNumber)) {
+      throw new DuplicateClaimException(
+          String.format(
+              "A claim with line number %d already exists for the submission.", lineNumber));
```
naseberry, line 147: nit: has the same http status and same logical failure as `DataClaimsExceptionHandler.java:60-61`, but they have two different detail strings.

claims-data/service/src/main/resources/db/migration/V45__add_unique_constraint_claim_submission_line_number.sql
vvartakMoJ: Yes, I agree, if we do this with a down-time (no traffic) then the write lock should not affect anything.

claims-data/service/src/main/resources/application.yml
```diff
@@ -20,6 +20,12 @@ spring:
       reporting_password: "reporting_password"
       data_factory_username: ${DATA_FACTORY_USERNAME:data_factory_user}
       data_factory_password: ${DATA_FACTORY_PASSWORD:data_factory_password}
+      # Cutoff for the partial unique index on claim(submission_id, line_number) added in V45.
+      # Defaults to the epoch so fresh/clean databases enforce uniqueness for ALL rows. Environments
+      # that already contain duplicate rows MUST override this (via CLAIM_LINE_NUMBER_UNIQUENESS_CUTOFF)
+      # with a timestamp after which the data is known to be clean, to grandfather the historical
+      # duplicates. See docs/release-notes/unique-claim-submission-line-number.md.
+      claim_line_number_uniqueness_cutoff: ${CLAIM_LINE_NUMBER_UNIQUENESS_CUTOFF:1970-01-01 00:00:00+00}
```
vvartakMoJ, line 30: What's the intended mechanism to provide this value for the environments? Do we need to add it to the values files or read it from a secret (via deployment.yaml, like the other values like DATA_FACTORY_USERNAME are provided)?

BenJusticeDigital: This was up for discussion as mentioned the other day we need a way to deploy and not hit any potential duplicates. environment specific environment variables is one so we can set them prior to deployment and not have to change the coded values.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/ClaimService.java
BenJusticeDigital: I agree that they are similar, but they cover different scenarios and provide different levels of diagnostic information.

In `ClaimService`, we already know the line number that caused the issue, which is valuable context and was identified through manual database verification. The other check acts as a generic safeguard and relies on an exception being thrown by the database. In that case, we do not have access to the specific line number because the exception only occurs if a race condition allows the operation to pass the earlier database validation.

So while both checks are intended to prevent the same overall problem, they are not equivalent and serve different purposes in terms of error handling and troubleshooting.

vvartakMoJ:
Closing PR so as to drop the ephemeral DB and start with an empty DB and run the changed migration V45 again.

claims-data/service/src/integrationTest/java/uk/gov/justice/laa/dstew/payments/claimsdata/controller/claim/amendments/AbstractAmendmentPatchIntegrationTest.java
```diff
@@ -61,6 +61,10 @@ abstract class AbstractAmendmentPatchIntegrationTest extends MockServerIntegrati
   // cache (keyed on office) isolated between tests and classes.
   private static final AtomicInteger OFFICE_SEQ = new AtomicInteger();
 
+  // Static, monotonically increasing line numbers so multiple claims created within the same
+  // submission never collide on the uq_claim_submission_line_number unique constraint.
+  private static final AtomicInteger LINE_SEQ = new AtomicInteger();
```
siddharthshroff-MoJ, line 66: Can this pose a risk when tests run in parallel ? I understand it is Atomic but having it static concerns me

claims-data/service/src/main/resources/application.yml
vvartakMoJ: I have now added a commit to supply "now()" as the value for this setting, instead of having to create secrets in each environment for this purpose.

claims-data/service/src/integrationTest/java/uk/gov/justice/laa/dstew/payments/claimsdata/controller/claim/amendments/AbstractAmendmentPatchIntegrationTest.java
vvartakMoJ: I think static atomic is in fact better than instance for these reasons:

• AtomicInteger.incrementAndGet() is thread-safe: even under parallel execution, no two callers ever receive the same value. The uniqueness invariant the field exists to protect holds regardless of parallelism.
• Being static actually strengthens parallel-safety here, not weakens it. A single JVM-wide monotonic sequence guarantees uniqueness across every subclass and thread. If it were an instance field instead, two test instances running in parallel would each start from 0 and could generate colliding line numbers when writing into the same submission — reintroducing exactly the constraint violation this guards against. The same reasoning is why OFFICE_SEQ (line 62) is deliberately static, as its comment explains.
• The values are only required to be distinct, never a specific number, so there are no order/value assertions that shared mutable state could disturb.
The only theoretical "downside" of a static counter is that it never resets and grows for the JVM's lifetime — irrelevant here (int won't overflow within a test run, and no reset is needed).

siddharthshroff-MoJ (approved):
LGTM, approving after rebasing

naseberry (approved):

PR #413, opened 2026-07-30
fix(DSTEW-1929):  Enforce unique live submission per office / area of law / period (409 on conflict).

vvartakMoJ:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/DSTEW-1929)

## Prevent duplicate live submissions (409 Conflict)

Enforces that only one **live** submission can exist for a given `office_account_number`, `area_of_law` and `submission_period`, and surfaces violations to clients as **409 Conflict** instead of a generic 500. "Live" excludes `VALIDATION_FAILED` and `REPLACED`, so a failed or replaced submission never blocks a fresh attempt for the same combination.

Enforcement is layered, mirroring the approach already used for the claim line-number uniqueness constraint:

- **Application pre-check (primary/common path)** — `SubmissionService.createSubmission` calls `requireNoConflictingLiveSubmission`, which uses a new repository method `existsByOfficeAccountNumberAndAreaOfLawAndSubmissionPeriodAndStatusNotIn` to fail fast with a clean `DuplicateSubmissionException` (409) before any write. Skipped when any keying field is null.
- **Database partial unique index (race-safe back-stop)** — catches anything that races past the pre-check.

### Changes

- **Migration V46** — partial unique index `uq_submission_live_office_aol_period` on `(office_account_number, area_of_law, submission_period) WHERE status NOT IN ('VALIDATION_FAILED','REPLACED')`. The cut-over is a Flyway placeholder `${live_submission_uniqueness_cutoff}` rather than a hard-coded date.
- **`DuplicateSubmissionException`** — `ClaimsDataException` carrying `HttpStatus.CONFLICT`; reuses the existing handler to produce an RFC 9457 ProblemDetail.
- **`SubmissionRepository`** — adds the `existsBy…StatusNotIn` derived query backing the pre-check.
- **`SubmissionService`** — adds the `requireNoConflictingLiveSubmission` guard (with `NON_LIVE_STATUSES` mirroring the index predicate) ahead of `save`.
- **`DataClaimsExceptionHandler`** — generalises the constraint mapping into a reusable `violatesConstraint(ex, constraintName)` and maps `uq_submission_live_office_aol_period` violations to 409 with a user-facing message (no raw SQL/constraint detail leaked). The existing claim line-number mapping now shares this path.
- **Config/deploy** — `application.yml` defaults `live_submission_uniqueness_cutoff` to the epoch (`1970-01-01…`) for local/CI so fresh DBs enforce uniqueness for **all** rows; the Helm `deployment.yaml` sets `LIVE_SUBMISSION_UNIQUENESS_CUTOFF` to deploy time so rows existing at deploy are grandfathered automatically.
- **Tests**
  - Unit — `SubmissionServiceTest`: `duplicateLiveSubmissionThrowsConflict`, `noConflictingLiveSubmissionPersists`, `missingKeyFieldSkipsDuplicateCheck`. `DataClaimsExceptionHandlerTest`: `handleDataIntegrityViolation_duplicateLiveSubmission_returnsConflict`.
  - Integration — `SubmissionControllerIntegrationTest`: 409 on a duplicate live submission (`postSubmission_shouldReturnConflict_WhenDuplicateLiveSubmission`) and 201 when the existing row is `REPLACED` (`postSubmission_shouldCreate_WhenExistingDuplicateIsReplaced`).

### Notes

- The DB index is a back-stop; primary enforcement lives in the application (and the companion project).
- The only Flyway migration adds this unique index, which doesn't affect logical replication, so it needs no coordination with the GLAD team.
- **No fixed release deadline.** The cut-over is per-environment via the `${live_submission_uniqueness_cutoff}` placeholder — Helm grandfathers all rows existing at deploy time — so there's no hard "must ship before date X" requirement and no risk of the index build failing on legacy duplicates.
- The cutoff literal is a `TIMESTAMPTZ` (not `DATE`) deliberately — a date comparison needs a timezone-dependent cast that PostgreSQL rejects in an index predicate (must be `IMMUTABLE`).## Checklist

- [x ] Tests should be passing: `./gradlew test`
- [x ] Github should not be reporting conflicts; you should have recently run `git rebase main`.
- [x ] Avoid mixing whitespace changes with code changes in the same commit. These make diffs harder to read and conflicts more likely.
- [x ] You should have looked at the diff against main and ensured that nothing unexpected is included in your changes.
- [x ] You should have checked that the commit messages say why the change was made.


claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/SubmissionService.java
```diff
@@ -97,7 +101,7 @@ public UUID createSubmission(SubmissionPost submissionPost) {
       }
     }
 
-    submissionRepository.save(submission);
+    persistSubmission(submission);
```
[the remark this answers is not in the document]
vvartakMoJ: The only client is our own events service so this is acceptable.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/service/SubmissionService.java
```diff
@@ -106,6 +110,40 @@ public UUID createSubmission(SubmissionPost submissionPost) {
...
+      }
+      throw ex;
+    }
+  }
+
+  private boolean isLiveSubmissionUniqueViolation(DataIntegrityViolationException ex) {
+    Throwable rootCause = ex.getMostSpecificCause();
+    return rootCause.getMessage() != null
+        && rootCause.getMessage().contains(LIVE_SUBMISSION_UNIQUE_INDEX);
+  }
```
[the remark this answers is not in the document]
vvartakMoJ: Changed to use Hibernate's ConstraintViolationException#getConstraintName() while keeping it driver-agnostic.

claims-data/service/src/test/java/uk/gov/justice/laa/dstew/payments/claimsdata/exception/DataClaimsExceptionHandlerTest.java
```diff
@@ -59,9 +59,9 @@ void handleGenericException_returnsProblemDetailWithInternalServerErrorStatus()
     assertThat(result.getBody().getTitle()).isEqualTo("Internal Server Error");
     assertThat(result.getBody().getDetail())
         .isEqualTo("An unexpected application error has occurred.");
-    assertThat(result.getBody().getType().toString())
-        .isEqualTo("https://claimsdata.payments.laa.justice.gov.uk/errors/runtime");
-    assertThat(result.getBody().getInstance().toString()).isEqualTo(TEST_REQUEST_URI);
+    assertThat(result.getBody().getType())
```
vvartakMoJ, line 62: These changes (`.hasToString`) are just to reduce the sonar qube warnings.

.helm/data-claims-api/templates/deployment.yaml
```diff
@@ -107,6 +107,8 @@ spec:
                   key: data-factory-password
             - name: CLAIM_LINE_NUMBER_UNIQUENESS_CUTOFF
               value: {{ dateInZone "2006-01-02 15:04:05-07:00" (now) "UTC" | quote }}
+            - name: LIVE_SUBMISSION_UNIQUENESS_CUTOFF
+              value: {{ dateInZone "2006-01-02 15:04:05-07:00" (now) "UTC" | quote }}
```
BenJusticeDigital, line 111: I assume we are setting these in the env vars so we don't get conflicts, i was not here for the last deploy so did not see the release process.

claims-data/service/src/integrationTest/java/uk/gov/justice/laa/dstew/payments/claimsdata/controller/SubmissionControllerIntegrationTest.java
```diff
@@ -592,6 +599,98 @@ void postNilSubmission_shouldReturnBadRequestWithValidationErrorDetails_WhenDupl
         .andReturn();
   }
 
+  @Test
+  @DisplayName(
+      "Should return 409 Conflict when a second live CREATED submission duplicates the office, area of law and period")
+  void postSubmission_shouldReturnConflict_WhenDuplicateLiveSubmission() throws Exception {
```
BenJusticeDigital, line 605: Let's convert this into a parameterized test and cover every known status value to ensure the behavior is validated consistently across all cases.

claims-data/service/src/integrationTest/java/uk/gov/justice/laa/dstew/payments/claimsdata/controller/AbstractIntegrationTest.java
```diff
@@ -1003,10 +1003,13 @@ protected UUID createMediationExportData(String officeAccountNumber) {
...
   protected Submission createIsolatedSubmission() {
+    UUID submissionId = Uuid7.timeBasedUuid();
     Submission submission =
         Submission.builder()
-            .id(Uuid7.timeBasedUuid())
-            .officeAccountNumber("totals-office")
+            .id(submissionId)
+            // Keep the office unique per submission so each is genuinely isolated and does not trip
+            // the uq_submission_live_office_aol_period unique index when a test creates several.
+            .officeAccountNumber("totals-office-" + submissionId)
```
BenJusticeDigital: Since this affects all tests, it would be better to override the behavior in the specific test that needs it instead of applying it globally.

claims-data/service/src/integrationTest/java/uk/gov/justice/laa/dstew/payments/claimsdata/repository/SubmissionRepositoryIntegrationTest.java
```diff
@@ -711,6 +711,8 @@ void responseShapeIsUnchangedOnlyTotalDiffers() {
           .usingRecursiveComparison()
           .ignoringFields(
               "submissionId", // Ignore root ID
+              "officeAccountNumber", // Unique per isolated submission (see
+              // createIsolatedSubmission)
```
BenJusticeDigital: I don't think we should relax this validation. It was added to ensure the Office Account Number is returned correctly, and this change effectively bypasses that check. Let's keep using valid office values and apply overrides only where they're needed.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/exception/DataClaimsExceptionHandler.java
```diff
@@ -234,46 +248,55 @@ public ResponseEntity<ProblemDetail> handleDatabaseOptimisticLockException(
...
       DataIntegrityViolationException ex, HttpServletRequest request) {
 
-    if (isDuplicateClaimLineNumberViolation(ex)) {
+    if (violatesConstraint(ex, UNIQUE_CLAIM_LINE_NUMBER_CONSTRAINT)) {
       log.warn("Duplicate claim line number within a submission:", ex.getMostSpecificCause());
       return buildProblemDetailResponse(
           HttpStatus.CONFLICT, DUPLICATE_CLAIM_LINE_NUMBER_MESSAGE, ex.getClass(), request);
     }
 
+    if (violatesConstraint(ex, UNIQUE_LIVE_SUBMISSION_CONSTRAINT)) {
```
BenJusticeDigital, line 271: This looks like a good candidate for refactoring. We could reduce the duplication by introducing a single testable method (e.g. validateConstraintAndError(...)) and parameterising the parts that differ.

BenJusticeDigital (changes requested):

.helm/data-claims-api/templates/deployment.yaml
vvartakMoJ: Hi Ben, this sets the value to the current date (when the deployment takes place in each env). It means we don't need a new secret value for each of these. They can be seen in the relevant postgres databases by querying the dictionary views.

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/exception/DataClaimsExceptionHandler.java
vvartakMoJ: Given that the return statement has to stay in this method, the refactoring does not reduce duplication much and makes the code harder to read (I tried a couple of different ways both looked worse than this).

vvartakMoJ:
Closing PR to destroy the ephemeral DB and restart (will reopen after it's gone).

BenJusticeDigital (approved):
Looks good. lets get QA to test thouroughly.

PR #421, opened 2026-08-06
chore(DSTEW-2085): Gracefully handle bad dates in mapper

BenJusticeDigital:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/DSTEW-2085)

This change strengthens amendment endpoint resilience by expanding test coverage for malformed and unexpected patch data.

### What changed
Added a new suite of parameterized MockMvc tests that submit invalid data across amendment patch fields.
Verified that invalid input is handled gracefully and does not result in an unhandled 500 Internal Server Error.
Added specific coverage for date fields using invalid date formats and invalid calendar dates.
Updated the amendment mapping layer to correctly handle date parsing failures and return an appropriate client error response rather than allowing exceptions to propagate as a 500.
Defects identified and fixed

### 1. Invalid dates returning HTTP 500

The new tests identified scenarios where malformed date values could cause exceptions during mapping, resulting in a 500 Internal Server Error.

The mapper has been updated to catch and handle date parsing failures, ensuring invalid input is reported as a controlled client error.

### 2. Date parser leniency bug

During testing we also discovered that the date parser was operating in a lenient manner and silently corrected invalid dates instead of rejecting them.

For example:
`31/02/2025`

was being interpreted as:
`28/02/2025`

rather than being treated as invalid input.

The parser has been corrected to use strict date validation so that invalid calendar dates are now rejected and reported appropriately.

### Benefits
- Improved API robustness when handling malformed amendment requests.
- Prevents invalid date values from being silently altered.
- Ensures clients receive consistent and meaningful error responses.
- Reduces the risk of unhandled exceptions reaching production.
- Increases confidence through broader negative-path and error-handling test coverage.

## Checklist


siddharthshroff-MoJ (approved):
LGTM

vvartakMoJ (approved):
LGTM!

PR #429, opened 2026-08-13
fix(DSTEW-2140): Map isAmended is claim responses

mboyd-moj:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/DSTEW-2140)

isAmended need to be manually mapped in MapStruct because its getter does not follow a format that MapStruct can automatically map (i.e. it has isAmended as it's getting but MapStruct would need getIsAmended)

Also bumped validation package version as now that the value is populated it needs to be ignored in claim schema validation

> **PR title must follow [Conventional Commits](https://www.conventionalcommits.org/)** (e.g. `feat(DSTEW-XXX): add outcome validation`).\
> It is linted in CI and drives the release version bump.\
> Allowed types: `feat`, `fix`, `chore`, `refactor`, `test`, `spike`, `docs`, `ci`, `infra`.\
> Version bump: `feat` → minor; any type followed by `!` or a `BREAKING CHANGE` footer → major; every other type (`fix`, `chore`, `refactor`, `test`, `spike`, `docs`, `ci`, `infra`) → patch.

## Checklist

### DIFF
--- a/claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/mapper/ClaimMapper.java
+++ b/claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/mapper/ClaimMapper.java
@@ -51,12 +51,14 @@ public interface ClaimMapper {
    */
   @Mapping(target = "isDutySolicitor", source = "dutySolicitor")
   @Mapping(target = "isYouthCourt", source = "youthCourt")
+  @Mapping(target = "isAmended", source = "amended")
   @Mapping(target = "submissionId", source = "submission.id")
   @Mapping(target = "submissionPeriod", source = "submission.submissionPeriod")
   ClaimResponse toClaimResponse(Claim entity);
 
   @Mapping(target = "isDutySolicitor", source = "dutySolicitor")
   @Mapping(target = "isYouthCourt", source = "youthCourt")
+  @Mapping(target = "isAmended", source = "amended")
   @Mapping(target = "submissionId", source = "submission.id")
   @Mapping(target = "submissionPeriod", source = "submission.submissionPeriod")
   @Mapping(target = "dateSubmitted", source = "submission.createdOn")
--- a/claims-data/service/build.gradle
+++ b/claims-data/service/build.gradle
@@ -77,8 +77,8 @@ dependencyManagement {
 
 dependencies {
     implementation project(':claims-data:api')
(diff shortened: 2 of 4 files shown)


claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/mapper/ClaimMapper.java
```diff
@@ -51,12 +51,14 @@ public interface ClaimMapper {
...
   @Mapping(target = "isDutySolicitor", source = "dutySolicitor")
   @Mapping(target = "isYouthCourt", source = "youthCourt")
+  @Mapping(target = "isAmended", source = "amended")
   @Mapping(target = "submissionId", source = "submission.id")
   @Mapping(target = "submissionPeriod", source = "submission.submissionPeriod")
   ClaimResponse toClaimResponse(Claim entity);
 
   @Mapping(target = "isDutySolicitor", source = "dutySolicitor")
   @Mapping(target = "isYouthCourt", source = "youthCourt")
+  @Mapping(target = "isAmended", source = "amended")
```
g-georgieva, line 61: Why are we mapping isAmended twice?

BenJusticeDigital: One is claim response and the other is claim response v2

BenJusticeDigital (commented):
Do the PACT tests need updating as part of this change?

If the contract, endpoint behaviour, or response structure has been modified, it would be good to make sure the corresponding PACT tests are updated accordingly so they continue to validate the expected consumer/provider contract.

mboyd-moj:
> Do the PACT tests need updating as part of this change?
> 
> If the contract, endpoint behaviour, or response structure has been modified, it would be good to make sure the corresponding PACT tests are updated accordingly so they continue to validate the expected consumer/provider contract.

@BenJusticeDigital Claim.isAmended in a primitive so we're already populating it for the producer side. I can update the pacts on consumer side (amend/submit) to check it though as at the moment that is missing

BenJusticeDigital (approved):

claims-data/service/src/main/java/uk/gov/justice/laa/dstew/payments/claimsdata/mapper/ClaimMapper.java
g-georgieva: Apologies, I did not see

g-georgieva (approved):

PR #386, opened 2026-07-15
feat(DSTEW-1645):  claim-history read contract by claim ID (SUBMISSION timeline)

BenJusticeDigital:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/DSTEW-1645)

### What & why
Delivers the claim-history read contract: `GET /api/v1/claims/{claimId}/history` returning a chronological timeline of common-envelope events. Ships with the `SUBMISSION` event populated from real submission data so the contract is proven end-to-end. The envelope, event-type enum, ordering and single `UNION ALL` read model ; this PR settles the contract and adds the full test pyramid.

### Contract
- **Endpoint:** `GET /api/v1/claims/{claimId}/history`, `limit` query param (`1..200`, default `50`).
- **Envelope (every event):** `event_type`, `event_timestamp`, `actor_id`, `source_id`, open `metadata` container (the extension point for later event types).
- **Result:** `{ "claim_id", "events": [...] }`.
- **Event types:** `SUBMISSION`, `AMENDMENT`, `ASSESSMENT`, `VOID` (only `SUBMISSION` populated here; other branches wired for follow-ups DSTEW-1812/1813/1814/1815).
- **Ordering:** `event_timestamp DESC`, deterministic `source_id DESC` (UUIDv7) tie-break.
- **Actor fallback:** `actor_id` always populated; `SYSTEM` where source has no user id.
- **Not found:** unknown claim id → `404`, consistent with existing claim reads.

### Changes
- **Spec:** array `content` → `events`; added `maximum: 200`; documented the one-response call pattern, ordering, fallback and 404.
- **Service:** existence check via `ClaimRepository`, throwing the existing `ClaimNotFoundException` (reuses shared error constant — no new exception).
- **Controller:** null-safe `limit`, builds `events`.
- **Docs:** `claim-history-timeline-poc.md` updated to the finalised contract.

### Tests
- **Unit:** service (not-found, delegation, page size), controller (envelope, metadata, actor fallback, 404, null-limit), row mapper.
- **Integration (Testcontainers):** submission-only timeline + 404; SUBMISSION mapping, `SYSTEM` fallback, same-timestamp tie-break at repository level.
- **Contract (Pact provider):** claim-history provider states + fixtures added; all 50 existing interactions still pass.

## Checklist

Before you ask people to review this PR please ensure the following and mark as complete when done:

- [x] Tests should be passing: `./gradlew test`
- [ ] Github should not be reporting conflicts; you should have recently run `git rebase main`.
[...]

PR #432, opened 2026-08-13
test(DSTEW-1645): BDD scenarios for claim history timeline parent-level guarantees

MrRazaH3:
## What

Adds Cucumber BDD coverage for the **cross-cutting parent-level guarantees** on `GET /api/v1/claims/{claimId}/history` (DSTEW-1645 parent) that no single child ticket (1811 / 1812 / 1813 / 1814 / 1815) can prove on its own.

## Scope

Read-side only. No production code changes — all additions live under `claims-data/service/src/bddTest`.

## ⚠️ De-scoped scenarios (audit)

**Two scenarios from the source branch could NOT be enabled against `main`.** Both surfaced during delivery-verification analysis. Documented inline in the feature-file de-scope banner.

| Tag | One-line reason | Blocks |
|---|---|---|
| `@DS1645_6` (all 4 Outline examples) | Actor fallback (`COALESCE(..., 'SYSTEM')`) unreachable across every event type. Every source `created_by_user_id` column feeding `HISTORY_SQL` is `TEXT NOT NULL`: `claim.created_by_user_id` (V3), `claim_amendment.created_by_user_id` (V39), `assessment.created_by_user_id` (V33). Postgres blocks null inserts everywhere — the fallback is defensive dead code across all 4 event types. Same root cause as `@DS1811_3` in #431 but broader. | Product decision: close as unreachable (recommended — a single decision covers all 4 event types and `@DS1811_3`). |
| `@DS1645_7` | Write-to-read smoke through the real DSTEW-1593 amendment endpoint — needs the **write-side amendment BDD harness** (WireMock for PDA / FSP / metadata reference-data, feature-flag manipulation, event-service coordination). None of that infrastructure exists in `bddTest` today. Scenario body preserved verbatim in the feature file so it can be re-enabled once the harness lands. | Team decision on write-side harness ownership. Same block as DSTEW-1770 and DSTEW-2051's write-side scenarios. |

## Scenarios (`@dstew-1645`)

| Tag | Purpose |
|---|---|
| `@DS1645_1` | Mixed timeline — SUBMISSION + AMENDMENT ×2 + ASSESSMENT + VOID all present, ordered by `event_timestamp DESC, source_id DESC`, no unknown event types slip through. |
| `@DS1645_2` | Failed amendment attempts (eligibility gate / PDA / FSP / final-save-guard) leave no `claim_amendment` row, so the timeline surfaces exactly the count of committed attempts. |
| `@DS1645_3` | New Matter Starts rows on the parent submission do NOT appear as timeline events (no branch in `HISTORY_SQL`) and no NMS-shaped metadata leaks into events. |
| `@DS1645_4` | AMENDMENT event never leaks the raw `request_payload` or the full `before_state` snapshot — only the versioned `metadata.changes[]` from `diff` is exposed. Sentinel markers in the row make a leak trivially detectable. |
| `@DS1645_5` | Submission-only / minimal history returns exactly one SUBMISSION event with the documented `{claim_id, events}` envelope. |

## Structural changes

- **`BddStepFailures.step(context, body)`** wrapper — same standing-rule helper added by #428 / #430 / #431. Will merge cleanly with any of them.
- **`BddApiStepSupport.getClaimHistory(UUID)`** — HTTP helper.
- **`BddTestConstants.GET_CLAIM_HISTORY_PATH`** — path constant.
- **`BddHooks`** — adds `claimAmendmentRepository.deleteAll()` before `claimRepository.deleteAll()` (FK cleanup ordering — same fix as in #428).

### Seeding notes

- `Submission → Claim → ClaimSummaryFee → ClaimAmendment / Assessment / MatterStart` seeded via JPA.
- `created_on` values on `claim`, `claim_amendment`, `assessment` pinned via native SQL UPDATE (bypasses `@CreationTimestamp` + `Claim.createdOn` being `@Column(updatable = false)` per the DSTEW-2051 fix).
- `OffsetDateTime.ofInstant(instant, ZoneOffset.UTC)` at the JDBC boundary — Postgres driver rejects raw `java.time.Instant`.
- Leak markers `PAYLOAD_LEAK_MARKER_PARENT_1645` / `BEFORE_STATE_LEAK_MARKER_PARENT_1645` injected into `claim_amendment.request_payload` and `.before_state` in `@DS1645_4` so any leak is caught by a single `assertThat(event.toString()).doesNotContain(marker)`.

## Local vs UAT

Local-only. Seeded through JPA + native SQL in step logic; asserted against the live REST endpoint via `TestRestTemplate`. No UAT hooks.

## Endpoints exercised

- `GET /api/v1/claims/{claimId}/history`

## Verification

- `./gradlew :claims-data:service:spotlessCheck` — clean.
- `./gradlew :claims-data:service:bddTest -Dcucumber.filter.tags=@dstew-1645` — **5/5 in-scope pass**.
- `./gradlew :claims-data:service:bddTest` — **81/81 pass** (full BDD suite, no regressions).

## Checklist

- [x] Story confirmed delivered on `main` (`f24734a7 feat(DSTEW-1645): claim-history read contract by claim ID (SUBMISSION timeline) (#386)` for the read-model baseline; children 1812/1813/1814/1815 layered the AMENDMENT/ASSESSMENT/VOID branches on top).
- [x] Feature file copied 1:1 from `DSTEW-1999-amend-claims-bdd-scenarios` and adjusted only to remove undeliverable scenarios (banner + PR audit block document the reasons).
- [x] All step methods use the new `BddStepFailures.step(...)` wrapper.
- [x] No production code changed.
- [x] No unrelated files staged.

## Jira

- DSTEW-1645

### DIFF
--- a/claims-data/service/src/bddTest/java/uk/gov/justice/laa/dstew/payments/claimsdata/bdd/context/BddResponseContext.java
+++ b/claims-data/service/src/bddTest/java/uk/gov/justice/laa/dstew/payments/claimsdata/bdd/context/BddResponseContext.java
@@ -1,13 +1,19 @@
 package uk.gov.justice.laa.dstew.payments.claimsdata.bdd.context;
 
+import com.fasterxml.jackson.databind.JsonNode;
 import lombok.Getter;
 import lombok.Setter;
 
-/** Shared response state for BDD contexts. */
+/** Shared JSON response state for BDD contexts. */
 @Getter
 @Setter
 public abstract class BddResponseContext {
 
   private Integer lastStatusCode;
-  private String lastResponseBody;
+  private JsonNode lastResponseBody;
+
+  public void clear() {
+    lastStatusCode = null;
+    lastResponseBody = null;
+  }
 }
--- a/claims-data/service/src/bddTest/java/uk/gov/justice/laa/dstew/payments/claimsdata/bdd/context/BddResponseContext.java
+++ b/claims-data/service/src/bddTest/java/uk/gov/justice/laa/dstew/payments/claimsdata/bdd/context/BddResponseContext.java
@@ -1,0 +1,13 @@
+package uk.gov.justice.laa.dstew.payments.claimsdata.bdd.context;
+
+import lombok.Getter;
+import lombok.Setter;
+
+/** Shared response state for BDD contexts. */
+@Getter
+@Setter
+public abstract class BddResponseContext {
+
+  private Integer lastStatusCode;
+  private String lastResponseBody;
+}
--- a/claims-data/service/src/bddTest/java/uk/gov/justice/laa/dstew/payments/claimsdata/bdd/context/ClaimHistoryContext.java
+++ b/claims-data/service/src/bddTest/java/uk/gov/justice/laa/dstew/payments/claimsdata/bdd/context/ClaimHistoryContext.java
@@ -1,6 +1,5 @@
 package uk.gov.justice.laa.dstew.payments.claimsdata.bdd.context;
 
-import com.fasterxml.jackson.databind.JsonNode;
 import io.cucumber.spring.ScenarioScope;
 import java.util.UUID;
 import lombok.Getter;
@@ -15,12 +14,9 @@ import org.springframework.stereotype.Component;
 public class ClaimHistoryContext extends BddResponseContext {
 
   private UUID currentClaimId;
-  private JsonNode lastResponse;
 
   public void reset() {
+    super.clear();
     currentClaimId = null;
-    lastResponse = null;
-    setLastStatusCode(null);
-    setLastResponseBody(null);
   }
 }
--- a/claims-data/service/src/bddTest/java/uk/gov/justice/laa/dstew/payments/claimsdata/bdd/context/ClaimHistoryContext.java
+++ b/claims-data/service/src/bddTest/java/uk/gov/justice/laa/dstew/payments/claimsdata/bdd/context/ClaimHistoryContext.java
@@ -1,0 +1,34 @@
+package uk.gov.justice.laa.dstew.payments.claimsdata.bdd.context;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import java.util.UUID;
+import org.springframework.stereotype.Component;
+
+/** Shared context for claim history BDD scenarios. */
+@Component
+public class ClaimHistoryContext {
+
+  private UUID currentClaimId;
+  private JsonNode lastResponse;
+
+  public UUID getCurrentClaimId() {
+    return currentClaimId;
+  }
+
+  public void setCurrentClaimId(UUID claimId) {
+    this.currentClaimId = claimId;
+  }
+
+  public JsonNode getLastResponse() {
+    return lastResponse;
+  }
+
+  public void setLastResponse(JsonNode response) {
+    this.lastResponse = response;
+  }
+
+  public void reset() {
+    this.currentClaimId = null;
+    this.lastResponse = null;
+  }
+}
--- a/claims-data/service/src/bddTest/java/uk/gov/justice/laa/dstew/payments/claimsdata/bdd/context/ClaimHistoryContext.java
+++ b/claims-data/service/src/bddTest/java/uk/gov/justice/laa/dstew/payments/claimsdata/bdd/context/ClaimHistoryContext.java
@@ -1,34 +1,26 @@
 package uk.gov.justice.laa.dstew.payments.claimsdata.bdd.context;
 
 import com.fasterxml.jackson.databind.JsonNode;
+import io.cucumber.spring.ScenarioScope;
 import java.util.UUID;
+import lombok.Getter;
+import lombok.Setter;
 import org.springframework.stereotype.Component;
 
 /** Shared context for claim history BDD scenarios. */
 @Component
-public class ClaimHistoryContext {
+@ScenarioScope
+@Getter
+@Setter
+public class ClaimHistoryContext extends BddResponseContext {
 
   private UUID currentClaimId;
   private JsonNode lastResponse;
 
-  public UUID getCurrentClaimId() {
-    return currentClaimId;
-  }
-
-  public void setCurrentClaimId(UUID claimId) {
-    this.currentClaimId = claimId;
-  }
-
-  public JsonNode getLastResponse() {
-    return lastResponse;
-  }
-
-  public void setLastResponse(JsonNode response) {
-    this.lastResponse = response;
-  }
-
   public void reset() {
-    this.currentClaimId = null;
-    this.lastResponse = null;
+    currentClaimId = null;
+    lastResponse = null;
+    setLastStatusCode(null);
+    setLastResponseBody(null);
   }
 }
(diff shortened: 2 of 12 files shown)


claims-data/service/src/bddTest/resources/features/bdd/claimHistoryTimelineParent.feature
```diff
@@ -0,0 +1,105 @@
...
+  #
+  # Following the DSTEW-1595 / DSTEW-1646 parent-file precedent, this file
+  # covers ONLY cross-cutting guarantees that no single child can prove:
+  #   * Mixed timeline (all 4 event types together, deterministic order).
+  #   * Failed amendment attempts have NO event (crosses the write path).
+  #   * New Matter Starts NOT in the timeline (submission-level exclusion).
+  #   * Raw request payload + full before-state NOT exposed in main response.
+  #   * Submission-only / minimal-history shape.
+  #   * Actor fallback when the source user id is missing.
+  #   * Write-to-read smoke through DSTEW-1593 (persisted, not seeded).
```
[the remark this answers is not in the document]
MrRazaH3: Fixed in 7ddf66a5.

Two changes:

1. Removed the overclaiming bullets from the "cross-cutting guarantees" list at the top of the file — the "Actor fallback", "Write-to-read smoke" and "Failed amendment attempts have NO event" bullets no longer appear on the list because the corresponding scenarios are de-scoped.

2. Deleted the pointer-to-commit block at the bottom and preserved `@DS1645_6` (Actor fallback Scenario Outline) and `@DS1645_7` (Write-to-read smoke) inline as commented-out Gherkin blocks. The scenarios are visible in this file exactly as they were authored — a reader working from this file alone can now see what they assert, and they can be un-commented once the delivery gap closes (actor-fallback resolution is not on `main`; write-to-read smoke needs the write-side amendment harness deferred to DSTEW-1770). Jira tickets will be raised to track those two follow-ups.

claims-data/service/src/bddTest/java/uk/gov/justice/laa/dstew/payments/claimsdata/bdd/steps/ClaimHistoryTimelineParentSteps.java
```diff
@@ -0,0 +1,642 @@
...
+        "seed a claim + only the successful amendment row (failed attempts leave no persisted row)",
+        () -> {
+          seedClaim();
+          List<Map<String, String>> rows = table.asMaps();
+          for (Map<String, String> row : rows) {
+            String outcome = row.get("outcome");
+            // Only the 'committed successfully' attempt creates a claim_amendment row on `main`.
+            // Failed attempts (eligibility, PDA, FSP, final-save-guard) short-circuit before the
+            // write, so seeding just the successful one exactly mirrors production.
+            if (outcome != null && outcome.toLowerCase().contains("committed")) {
```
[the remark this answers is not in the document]
MrRazaH3: Fixed in 7ddf66a5. Agreed — the step body iterated the outcome column but only inserted the `committed successfully` row, so a regression that started persisting eligibility/PDA/FSP/final-guard failures would never have been caught.

Per your suggestion, de-scoped `@DS1645_2` until the write-side amendment harness exists (deferred to DSTEW-1770). The scenario body is preserved as a commented-out block in the feature file so nothing is lost; the step definition (`aClaimExistsWithTheFollowingAmendmentAttempts`) remains in place so the scenario can be un-commented once the harness lands and can genuinely drive each failure path. A separate Jira ticket will be raised to track this.

siddharthshroff-MoJ (dismissed):
Made changes to implement shared context and common steps for claim history BDD scenarios. LGTM now

claims-data/service/src/bddTest/java/uk/gov/justice/laa/dstew/payments/claimsdata/bdd/context/ClaimHistoryContext.java
```diff
@@ -0,0 +1,34 @@
+package uk.gov.justice.laa.dstew.payments.claimsdata.bdd.context;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import java.util.UUID;
+import org.springframework.stereotype.Component;
+
+/** Shared context for claim history BDD scenarios. */
+@Component
```
BenJusticeDigital, line 10: We already use Lombok for generating getters and setters, and this feels very similar in concept to BddScenarioContext used in the claims-data service:

https://github.com/ministryofjustice/laa-data-claims-api/blob/main/claims-data/service/src/bddTest/java/uk/gov/justice/laa/dstew/payments/claimsdata/bdd/context/BddScenarioContext.java

Given the overlap, it might be worth considering a common base/context class that both implementations can extend. At a minimum, the following fields appear to be applicable across both use cases:

```
private int lastStatusCode;
private String lastResponseBody;
```

It would also be good to decide on the most appropriate type for the response body. If the response is always JSON, using a JsonNode could provide more flexibility and avoid repeated parsing. If the content can vary, keeping it as a String may be the simpler option.

This could help reduce duplication, keep the test context consistent across services, and make future enhancements easier to apply in a single place.

claims-data/service/src/bddTest/java/uk/gov/justice/laa/dstew/payments/claimsdata/bdd/steps/ClaimHistoryTimelineCommonSteps.java
```diff
@@ -0,0 +1,32 @@
...
+    step(
+        "GET /api/v1/claims/" + claimHistoryContext.getCurrentClaimId() + "/history",
+        () -> {
+          UUID claimId = claimHistoryContext.getCurrentClaimId();
+          if (claimId == null) {
+            throw new AssertionError(
+                "No claim id has been established yet — expected a prior Given step.");
+          }
+          JsonNode response = api.getClaimHistoryJson(claimId);
+          claimHistoryContext.setLastResponse(response);
```
BenJusticeDigital: I would also set the status code here if we go with the shared context/base class approach suggested above. If we're standardising the response handling, it may also be worth storing the response as a JsonNode rather than a String, assuming the API responses are always JSON. That would make assertions easier and avoid repeated parsing throughout the tests. If non-JSON responses are expected, then keeping it as a String would be more appropriate.

BenJusticeDigital (changes requested):

claims-data/service/src/bddTest/java/uk/gov/justice/laa/dstew/payments/claimsdata/bdd/steps/ClaimHistoryTimelineCommonSteps.java
siddharthshroff-MoJ: Good point — I’ve updated this to set the status code on the shared response context as well.

claims-data/service/src/bddTest/java/uk/gov/justice/laa/dstew/payments/claimsdata/bdd/context/ClaimHistoryContext.java
siddharthshroff-MoJ: Good shout! I ahve refactored both scenarios into parent BDDResponseBaseContext

claims-data/service/src/bddTest/java/uk/gov/justice/laa/dstew/payments/claimsdata/bdd/context/BddResponseContext.java
```diff
@@ -0,0 +1,13 @@
...
+import lombok.Getter;
+import lombok.Setter;
+
+/** Shared response state for BDD contexts. */
+@Getter
+@Setter
+public abstract class BddResponseContext {
+
+  private Integer lastStatusCode;
+  private String lastResponseBody;
```
BenJusticeDigital: Just a couple of points:

1. I think we should return either a string or JSON in the response body method, not both. For consistency, it would probably be better to always return JSON, since that's what we return in from the API.
2. For the clear() method, it might make sense to add a base implementation in the abstract class to clear the two fields it owns, then call it with super.clear() from subclasses. That would avoid having to clear those fields in every implementation, and if we add more fields to the abstract class later, we won't need to update multiple classes to handle the new variables.

BenJusticeDigital (changes requested):

BenJusticeDigital (approved):

PR #435, opened 2026-08-14
fix: Helm ServiceMonitor indentation

naseberry:
## What

CP s been noticing a lot of errors from prometheus, telling us it's scraping out of order timestamps.

The cause of that looks to be service monitors in the UAT namespace (due to the ephemeral environment) aren't selecting their corresponding service properly, so it's scraping much more than it should and ends up reporting duplicate metrics.

There was a missing indentation in the helm chart, as the labels aren't under `spec.selector.matchLabels`, they are inline.

Reference CP Issue: https://github.com/ministryofjustice/cloud-platform/issues/8452

## Checklist

### DIFF
--- a/.helm/data-claims-api/templates/serviceMonitor.yaml
+++ b/.helm/data-claims-api/templates/serviceMonitor.yaml
@@ -6,7 +6,7 @@ metadata:
 spec:
   selector:
     matchLabels:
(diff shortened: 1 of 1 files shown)


vvartakMoJ (approved):
LGTM!

PR #445, opened 2026-08-20
test(DSTEW-1774|DSTEW-1646): add step definitions and feature files for PDA outcome mapping and parent integration scenarios

siddharthshroff-MoJ:
## What

[Link to story-1](https://dsdmoj.atlassian.net/browse/DSTEW-1774)
[Link to story-2](https://dsdmoj.atlassian.net/browse/DSTEW-1646)

Add BDD test suites for `amendmentsPdaParentIntegration.feature` and `amendmentsPdaOutcomeMapping.feature`  covering PDA validation-message mapping, terminal-technical failure,  no-save orchestration, and safe-context logging

## Checklist

### DIFF
--- a/claims-data/service/src/bddTest/java/uk/gov/justice/laa/dstew/payments/claimsdata/bdd/steps/AmendmentPdaOutcomeMappingSteps.java
+++ b/claims-data/service/src/bddTest/java/uk/gov/justice/laa/dstew/payments/claimsdata/bdd/steps/AmendmentPdaOutcomeMappingSteps.java
@@ -1,0 +1,299 @@
+package uk.gov.justice.laa.dstew.payments.claimsdata.bdd.steps;
+
+import static uk.gov.justice.laa.dstew.payments.claimsdata.bdd.config.BddTestConstants.DEFAULT_OFFICE;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import io.cucumber.java.en.Given;
+import io.cucumber.java.en.Then;
+import java.time.LocalDate;
+import java.time.Month;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.beans.factory.annotation.Autowired;
+import uk.gov.justice.laa.dstew.payments.claimsdata.bdd.context.SharedAmendmentPatchContext;
+import uk.gov.justice.laa.dstew.payments.claimsdata.bdd.generator.SubmissionPeriodHelper;
+import uk.gov.justice.laa.dstew.payments.claimsdata.entity.Claim;
+import uk.gov.justice.laa.dstew.payments.claimsdata.entity.Submission;
+import uk.gov.justice.laa.dstew.payments.claimsdata.model.AreaOfLaw;
+import uk.gov.justice.laa.dstew.payments.claimsdata.model.ClaimStatus;
+import uk.gov.justice.laa.dstew.payments.claimsdata.model.SubmissionStatus;
+import uk.gov.justice.laa.dstew.payments.claimsdata.repository.ClaimRepository;
+import uk.gov.justice.laa.dstew.payments.claimsdata.repository.SubmissionRepository;
+import uk.gov.justice.laa.dstew.payments.claimsdata.util.Uuid7;
+
+/**
+ * Step definitions for {@code amendmentsPdaOutcomeMapping.feature} (DSTEW-1774) and shared step
+ * definitions co-used by {@code amendmentsPdaParentIntegration.feature} (DSTEW-1646).
+ *
+ * <p>The amendment-path PDA client and its outcome mapping (validation-message codes vs.
+ * terminal-technical failure) live in downstream stories not yet exposed to the BDD harness. The
+ * scenarios here therefore rely on existing amendment-metadata step glue for the {@code When} and
+ * for common {@code Then}s ({@code the amendment is rejected with the following errors}, {@code the
+ * endpoint responds with a controlled terminal failure "..."}, {@code no amendment state was
+ * committed}); PDA-specific behaviour is asserted as {@code log.info("[spec-guard] ...")} until
+ * DSTEW-1773/1774 wiring surfaces the outcome for observation. This class also owns the fixture
+ * steps ({@code an original claim exists with feeCode ... and officeCode ... and effectiveDate
+ * ...}, {@code an amendment updates the claim to feeCode ... and effectiveDate ...}, {@code the
+ * claim persisted state matches the pre-amendment state}) that are reused by the parent-integration
+ * feature (DSTEW-1646).
+ */
+@Slf4j
+public class AmendmentPdaOutcomeMappingSteps {
+
+  private static final String SEED_ACTOR = "bdd-DSTEW-1774";
+
+  @Autowired private SharedAmendmentPatchContext sharedPatchContext;
+  @Autowired private SubmissionPeriodHelper periodHelper;
+  @Autowired private ClaimRepository claimRepository;
+  @Autowired private SubmissionRepository submissionRepository;
+
+  private final ObjectMapper objectMapper = new ObjectMapper();
+
+  // ---------------------------------------------------------------------------
+  // Shared fixture steps (also used by DSTEW-1646 parent integration)
+  // ---------------------------------------------------------------------------
+
+  @Given(
+      "an original claim exists with feeCode {string} and officeCode {string} and effectiveDate"
+          + " {string}")
+  public void originalClaimExistsWithFeeOfficeAndEffectiveDate(
+      String feeCode, String officeCode, String effectiveDate) {
+    log.info(
+        "[spec-guard] Original claim fixture: feeCode={} officeCode={} effectiveDate={}",
+        feeCode,
+        officeCode,
+        effectiveDate);
+    provisionAmendableClaim(feeCode);
+  }
+
+  @Given("an amendment updates the claim to feeCode {string} and effectiveDate {string}")
+  public void amendmentUpdatesClaimToFeeCodeAndEffectiveDate(String feeCode, String effectiveDate) {
+    log.info("[spec-guard] Amendment intent: feeCode={} effectiveDate={}", feeCode, effectiveDate);
+    // Ensure a claim is provisioned even when this step runs before the "original claim" fixture
+    // (some scenarios in the parent-integration feature omit the "original claim" phrase). If
+    // already provisioned by an earlier step, keep the existing shared context and just push a
+    // fee-code delta into the patch body.
+    if (!sharedPatchContext.isPopulated()) {
+      provisionAmendableClaim("CAPA");
+    }
+    ObjectNode root = objectMapper.createObjectNode();
+    root.put("client_forename", "Amended");
+    root.put("fee_code", feeCode);
+    root.put("version", 0);
+    sharedPatchContext.setPatchJson(root.toString());
+  }
+
+  @Then("the claim persisted state matches the pre-amendment state")
+  public void claimPersistedStateMatchesPreAmendmentState() {
+    log.info(
+        "[spec-guard] Expected: persisted claim state unchanged from pre-amendment snapshot"
+            + " (amendment-path rollback / short-circuit)");
+  }
+
+  // ---------------------------------------------------------------------------
+  // Provisioning helper
+  // ---------------------------------------------------------------------------
+
+  private void provisionAmendableClaim(String feeCode) {
+    String office = DEFAULT_OFFICE;
+    String period = periodHelper.nextAvailablePeriod(office, AreaOfLaw.LEGAL_HELP);
+
+    Submission submission =
+        submissionRepository.saveAndFlush(
+            Submission.builder()
+                .id(Uuid7.timeBasedUuid())
+                .officeAccountNumber(office)
+                .submissionPeriod(period)
+                .areaOfLaw(AreaOfLaw.LEGAL_HELP)
+                .status(SubmissionStatus.CREATED)
+                .createdByUserId(SEED_ACTOR)
+                .providerUserId(SEED_ACTOR)
+                .createdOn(java.time.Instant.now())
+                .build());
+
+    Claim claim =
+        claimRepository.saveAndFlush(
+            Claim.builder()
+                .id(Uuid7.timeBasedUuid())
+                .submission(submission)
+                .status(ClaimStatus.VALID)
+                .feeCode(feeCode != null ? feeCode : "CAPA")
+                .lineNumber(1)
+                .matterTypeCode("MAT01")
+                .uniqueFileNumber("010725/001")
+                .caseReferenceNumber("CRN-1774")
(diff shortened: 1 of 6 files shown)


MrRazaH3 (commented):
Hi Sid — really nice work on this. The real MockServer + real socket timeouts pattern is a solid choice for the PDA outcome scenarios, especially the timeout tripping in seconds. A few things I'd love to align on before merge — none of these are "you got it wrong", more "let's make sure we don't collide with what's landing next to this."

## The big one — architectural overlap with the DSTEW-2301 harness (PR #455)

PR #455 is landing (or has just landed) a parallel amendment BDD harness that takes the **opposite** approach: it registers `@MockitoBean FeeSchemePlatformRestClient` and `@MockitoBean ValidationService` on `CucumberSpringConfiguration`, so external calls never leave the JVM. Your approach here uses the real client beans and stubs at the HTTP transport (MockServer). Both patterns are defensible, but they don't coexist — the `@MockitoBean` wins, and your MockServer stubs would be silently unused because the client bean is mocked before the real HTTP call is issued.

When both PRs are on `main`, one of these has to give. Options I can see:

1. **Rebase this onto `DSTEW-2301-bdd`** (or whatever the harness base ends up being) and drop the `@MockitoBean` for `FeeSchemePlatformRestClient` / `ValidationService` so your real-HTTP pattern wins on the amendment path.
2. **Tag-scope the mocks** — the harness `@MockitoBean` only fires on `@amendments-harness`-tagged scenarios; the real transport wins for `@pda`-tagged scenarios.
3. **Swap this PR to Mockito-baseline** and lose the real-timeout tripping (which would be a shame — that's the strongest signal you have here).

Might be worth 15 min with Peter / the DSTEW-2301 author to pick one before we end up unmerging things. Happy to help coordinate.

## The DSTEW-1646 spec-guard problem

`AmendmentPdaParentIntegrationSteps` is admirably honest about it — the class Javadoc says *"every step here is a spec-guard log.info entry"*. Which means both `@DS1646_1` (early-rejection short-circuit) and `@DS1646_2` (post-PDA rollback) currently pass by not asserting anything. On `@Regression` that's a false-green trap: any real regression in the ordering/atomicity guarantees will land in prod because the tagged scenarios all pass unconditionally.

Our current team rule (living in `memory.md`) is that scenarios we can't yet assert should be **commented out with a `# TODO DSTEW-XXXX` marker + a follow-up Jira ticket that (a) unblocks the assertion and (b) re-enables the scenario**. That way they surface in refinement/planning instead of hiding as green.

Two ways to fix it in this PR:
- Comment out `@DS1646_1` / `@DS1646_2` scenarios, add a follow-up ticket (I can help draft it — the observable hook they need is either a real short-circuit test-hook or a `SpyBean` on the orchestrator that lets us count PDA calls).
- **Or** make the log-only step methods `throw new PendingException()` (or `fail(...)`) — Cucumber will then mark the scenarios as pending/failed and they won't hide in green.

Left an inline note on that file too.

## Smaller stuff (inline)

- `PDA_READ_TIMEOUT_MS = 2000` in `CucumberSpringConfiguration` — thoughtful, but it applies to every BDD scenario, not just DSTEW-1773 / 1774. Any pre-existing non-timeout scenario that legitimately takes >2s to talk to PDA in the JVM-boot warm-up phase could flake.
- `"the amendment is rejected"` in `AmendmentPdaParentIntegrationSteps` — very generic phrase. When somebody adds a real amendment-rejection assertion elsewhere (this is very likely soon), we'll get `DuplicateStepDefinitionException`. PRs #451, #455, #458 all hit that cascade from more distinctive phrases than this one.
- `new ObjectMapper()` in `AmendmentPdaOutcomeMappingSteps.java:57` — Spring gives us a shared configured bean; `@Autowired` matches the rest of the codebase.
- `@Then("no amendment validation messages are returned alongside the terminal failure")` — the body only `log.info`'s the response, no assertion. Same silent-pass concern as the parent-integration file.
- `bddMockServerSupport.reset()` in `BddHooks#resetAll` (order = 0) — good spot. Just noting `BddAmendmentResetHook` from DSTEW-2301 runs at `order = -1`. If the two harnesses merge, the ordering plays nicely: yours runs after the mock-reset from the sibling. No change needed here — flagging so you're aware.

## Nice things I want to call out

- The class Javadoc on `AmendmentPdaOutcomeMappingSteps` is exemplary — explicitly names the tech-error code being asserted and cross-refs `ClaimAmendmentPdaCallIntegrationTest`. That's the standard.
- The `bdd.pda.newSubmissionReadTimeoutMs` fixture-only property is well-documented in the comment ("There is no separate production property today; this key is BDD-only"). Belt-and-braces would be adding a `TODO(DSTEW-XXXX)` next to it so we don't lose track once a real production separation appears, but honestly your comment is good enough.
- Real MockServer + real timeouts is genuinely a better signal than what the DSTEW-2301 harness gives you today. If the architectural conversation goes your way, this PR effectively upgrades the whole amendment BDD tier.

Overall — architecturally the biggest thing is deciding whose approach wins with #455. Once that's settled the rest is quick.

claims-data/service/src/bddTest/java/uk/gov/justice/laa/dstew/payments/claimsdata/bdd/CucumberSpringConfiguration.java
```diff
@@ -14,17 +20,77 @@
...
  */
 @CucumberContextConfiguration
 @ActiveProfiles("test")
 @SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
 @Import({AwsTestConfig.class, BddBeansConfiguration.class})
 public class CucumberSpringConfiguration {
 
+  /**
+   * Small enough to make amendment-path PDA timeout scenarios trip in seconds. Success scenarios
+   * ({@code @PDA_4}) must stay comfortably under this budget.
```
MrRazaH3: 2 seconds is quite tight for a global — it applies to every BDD scenario, not just the DSTEW-1773/1774 ones you added. If any pre-existing scenario happens to hit PDA on a cold JVM (Spring warm-up, first Testcontainers hit), it could flake with a spurious timeout. Two options:

1. Keep it here but scope the assertion by tag — only `@pda_timeout`-tagged scenarios pull this value.
2. Move it to a per-scenario `DynamicPropertySource` (or a `@BeforeTag` hook) that only trips it down for the timeout scenarios and restores default for the rest.

Not blocking for me — just flagging the blast radius.

claims-data/service/src/bddTest/java/uk/gov/justice/laa/dstew/payments/claimsdata/bdd/CucumberSpringConfiguration.java
```diff
@@ -14,17 +20,77 @@
...
+
+  /**
+   * Points the claims-validation-core external URLs at the running MockServer and shortens the
+   * amendment-path PDA read timeout so DSTEW-1773 / DSTEW-1774 timeout scenarios can trip in
+   * seconds.
+   */
+  @DynamicPropertySource
+  static void validatorProperties(DynamicPropertyRegistry registry) {
+    String baseUrl = MOCK_SERVER.getEndpoint();
+
```
MrRazaH3: This is the main architectural collision point I mentioned in the summary — pointing external URLs at MockServer only works if the real `FeeSchemePlatformRestClient` and `ValidationService` beans are wired in. DSTEW-2301 (PR #455) is landing `@MockitoBean` for both of those on this same class, which will make your MockServer stubs unused because the client bean is mocked at the JVM level.

Worth coordinating with the DSTEW-2301 owner on which approach the amendment BDD suite should standardise on.

claims-data/service/src/bddTest/java/uk/gov/justice/laa/dstew/payments/claimsdata/bdd/steps/AmendmentPdaParentIntegrationSteps.java
```diff
@@ -0,0 +1,76 @@
...
+ * harness today, so every step here is a spec-guard {@code log.info} entry.
+ *
+ * <p>Shared fixture / assertion steps ({@code an original claim exists with feeCode ... and
+ * officeCode ... and effectiveDate ...}, {@code an amendment updates the claim to feeCode ... and
+ * effectiveDate ...}, {@code the claim persisted state matches the pre-amendment state}) are owned
+ * by {@link AmendmentPdaOutcomeMappingSteps} to avoid {@code DuplicateStepDefinitionException} and
+ * are picked up by cucumber's shared glue.
+ */
+@Slf4j
+public class AmendmentPdaParentIntegrationSteps {
```
MrRazaH3: Really appreciate the honesty in this class Javadoc — but a `@Regression`-tagged scenario that passes without asserting anything is exactly the false-green trap our team memory says we should avoid.

Cleanest fix: either

- Comment out `@DS1646_1` and `@DS1646_2` in the feature file with `# TODO DSTEW-XXXX` and raise a follow-up ticket for the orchestration hooks these scenarios need, or
- Make the log-only step bodies `throw new PendingException()` — Cucumber will mark the scenarios as pending, they won't hide in green.

Either signals to future us that this coverage isn't real yet.

claims-data/service/src/bddTest/java/uk/gov/justice/laa/dstew/payments/claimsdata/bdd/steps/AmendmentPdaParentIntegrationSteps.java
```diff
@@ -0,0 +1,76 @@
...
+    log.info(
+        "[spec-guard] Fixture: amendment persistence engineered to fail AFTER PDA has returned"
+            + " success — expected to trigger atomic rollback");
+  }
+
+  // ---------------------------------------------------------------------------
+  // Then — parent-integration assertions
+  // ---------------------------------------------------------------------------
+
+  @Then("the amendment is rejected")
```
MrRazaH3: Really generic phrase — going to collide the moment another step class defines a real assertion for "amendment is rejected" (which is inevitable). We've hit `DuplicateStepDefinitionException` cascades on PRs #451, #455 and #458 from more distinctive phrases than this one, and the cascade masquerades as a hundred `UndefinedStepException`s.

Suggest something disambiguated like `"the amendment is rejected (parent-integration ordering guarantee)"` or `"the amendment is rejected via the pre-PDA short-circuit"`, so any real assertion elsewhere can keep the shortest phrase for itself.

claims-data/service/src/bddTest/java/uk/gov/justice/laa/dstew/payments/claimsdata/bdd/steps/AmendmentPdaOutcomeMappingSteps.java
```diff
@@ -0,0 +1,394 @@
...
+  private static final AtomicInteger OFFICE_SEQ = new AtomicInteger();
+
+  @Autowired private BddScenarioContext scenarioContext;
+  @Autowired private SharedAmendmentPatchContext sharedPatchContext;
+  @Autowired private SubmissionPeriodHelper periodHelper;
+  @Autowired private ClaimRepository claimRepository;
+  @Autowired private SubmissionRepository submissionRepository;
+  @Autowired private BddMockServerSupport mock;
+
+  private final ObjectMapper objectMapper = new ObjectMapper();
```
MrRazaH3: Small nit — Spring gives us a configured shared `ObjectMapper` bean; `@Autowired private ObjectMapper objectMapper;` matches what the other step classes in this codebase do and picks up any project-wide Jackson config (date formats, module registrations) for free.

claims-data/service/src/bddTest/java/uk/gov/justice/laa/dstew/payments/claimsdata/bdd/steps/AmendmentPdaOutcomeMappingSteps.java
```diff
@@ -0,0 +1,394 @@
...
+    boolean hasProblemDetail = body.path("status").isNumber();
+    assertThat(hasErrorsArray || hasProblemDetail)
+        .as(
+            "Response body must be a Step 12 multi-message envelope (errors array or"
+                + " ProblemDetail): %s",
+            body)
+        .isTrue();
+  }
+
+  @Then("no amendment validation messages are returned alongside the terminal failure")
```
MrRazaH3: This is a `@Then` that only `log.info`'s — it passes for any response. Same silent-pass concern I flagged on the parent-integration file: on `@Regression` this can never surface a real regression.

If the assertion genuinely can't be made from the harness today, either mark the scenario `@pending`, comment out and TODO, or throw `PendingException` from the step body. Passing green with no assertion is the false-green trap we're trying to avoid.

Same applies to `techLogContainsCorrelationId`, `techLogContainsPdaOutcome`, `techLogNoPayloadValues`, `techLogNoFinancialValues`, `pdaMonitoringRecords*` and `pdaMonitoringDoesNotContain*` below — they're all pure log.info spec-guards.

MrRazaH3 (approved):
Approving from a test-engineering ownership perspective — the real MockServer + real socket-timeout approach is the stronger long-term signal for the amendment BDD tier, and I'd rather unblock this so we can align PR #455 (DSTEW-2301) onto the same pattern than gate this on the architectural conversation.

To be transparent — the concerns I raised in my earlier review are **not fully resolved**. They're being carried into follow-up work rather than blocking merge:

### Immediate post-merge work (I'll own the tracking)

1. **DSTEW-2317** (already raised for PR #455 follow-up) will have its scope expanded to swap the `@MockitoBean FeeSchemePlatformRestClient` + `@MockitoBean ValidationService` on `CucumberSpringConfiguration` for the MockServer pattern this PR introduces, so the two harnesses converge. Biggest sub-risk: porting the DSTEW-1753 `@DS1753_8 @race` scenario, which today wires a `CyclicBarrier` inside `Mockito.doAnswer` on `validateClaim` — that has to move into a MockServer callback or a service test-hook.

2. **New follow-up ticket** — I'll raise a draft covering the residual concerns from this PR:
   - `AmendmentPdaParentIntegrationSteps` all-`log.info` scenarios (`@DS1646_1`, `@DS1646_2`) either commented out with Jira TODO or converted to `PendingException`, so they stop silently passing on `@Regression`.
   - Same treatment for the log-only `@Then` steps in `AmendmentPdaOutcomeMappingSteps` (`noValidationMessagesAlongsideTerminal`, `techLog*`, `pdaMonitoring*`).
   - Disambiguate the generic `"the amendment is rejected"` step phrase before it collides with a real amendment-rejection assertion.
   - Tag-scope the global `PDA_READ_TIMEOUT_MS = 2000` so it doesn't affect non-timeout scenarios.
   - Small nit: use `@Autowired ObjectMapper` instead of `new ObjectMapper()` in `AmendmentPdaOutcomeMappingSteps`.

Ticket number will be added here once raised.

Good work Sid — approving. Let's talk after merge about ordering the two follow-ups.

MrRazaH3 (approved):

siddharthshroff-MoJ:
Resolving the comments based on the discussed facts and long term approach of using real HTTP client calls using MockServer

